PR #4, opened 2021-07-15
Add actionlint to path when downloaded in a GitHub Action

dgholz:
This makes the Action environment better match the local environment, for tools which use `actionlint` locally and in a workflow.

I'm using [reviewdog](https://github.com/reviewdog/reviewdog), and I can configure `actionlint` to run the check locally in my `.reviewdog.yml`:
```
runner:
    actionlint:
    cmd: actionlint -no-color -oneline
    # actionlint is a Go program
    format: golint
```

but in an action, I need to specify the arguments & reviewdog format again:
```
      - name: Run actionlint
        env:
          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          ${{ steps.get_actionlint.outputs.executable }} -no-color -oneline | reviewdog -reporter=github-pr-review -fail-on-error -f golint
```

because the `cmd` part in the `.reviewdog.yml` assumes the `actionlint` binary is in the path. I can just run a separate step like:
```
      - name: Add actionlint to path
        run: |
          echo "${{ steps.get_actionlint.outputs.executable }}" >> "$GITHUB_PATH"
```

But it looks like it's simple enough to do it from the download script, and I think it's safe.

### DIFF
--- a/scripts/download-actionlint.bash
+++ b/scripts/download-actionlint.bash
@@ -74,4 +74,6 @@ echo "Done: $("${exe}" -version)"
 if [ -n "$GITHUB_ACTION" ]; then
     # On GitHub Actions, set executable path to output
     echo "::set-output name=executable::${exe}"
+    # and add it to the path
+    echo "${exe}" >> "$GITHUB_PATH"
 fi


rhysd:
Thank you for using actionlint and suggesting the idea.

However, adding the current directory to `$PATH` implicitly is dangerous since we cannot know what is put in the current directory. For example, if a file named `ls` exists in the current directory, `ls` command will no longer work in the user's workflow. And it is worse case that an untrusted binary is put in the current directory since it can be security issue. In my opinion, adding new directory to `$PATH` should be done by user explicitly.

dgholz:
Thanks for the quick response! Your explanation about the path makes sense, though truth be told I assumed the download script would write the executable to a new tempdir and didn't check if it wrote to the current directory.

> BTW, do you know reviewdog already officially supports actionlint?

I did not, I'll use that instead.

PR #5, opened 2021-07-16
Add support for Alpine linux

admiralAwkbar:
Hey! I'm trying to test and add this to the **Github-Super-Linter**
Needed to add some support for our Alpine based image

### DIFF
--- a/scripts/download-actionlint.bash
+++ b/scripts/download-actionlint.bash
@@ -17,6 +17,10 @@ case "$OSTYPE" in
         os=linux
         ext=tar.gz
     ;;
+    linux-musl*)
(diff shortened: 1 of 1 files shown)


rhysd (approved):

rhysd:
Thank you for catching it. It might be better to use `linux-*` for the `case` condition. I'll update it after merging this PR.

admiralAwkbar:
@rhysd  hey yeah that might work, wasnt sure how many versions on Linux out there so was going the smaller route :)

rhysd:
Actually we don't need to care about glibc or musl because Go does not rely on libc.

PR #15, opened 2021-07-26
cron schedule `*/5 * * * *` should be accepted

### LINKED ISSUE #14: About the lint behavior of scheduled job
When I run `actionlint`, the output is as follows:

```log
.github/workflows/schedule.yml:4:13: scheduled job runs too frequently. it runs once per 60 seconds [events]
  |
4 |     - cron: '* * * * *'
  |             ^~
```

According to the documentation at https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events, it seems that '* * * * *' appears to be the correct format.

Which is correct, GitHub Actions document or `actionlint` implementation?

shogo82148:
fixes https://github.com/rhysd/actionlint/issues/14#issuecomment-886716283
> It looks that 04696da warns */5 * * * *, however it's OK because the interval can be every 5 minutes.

### DIFF
--- a/rule_events.go
+++ b/rule_events.go
@@ -60,7 +60,7 @@ func (rule *RuleEvents) checkCron(spec *String) {
 	// (#14) https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events
 	//
 	// > The shortest interval you can run scheduled workflows is once every 5 minutes.
-	if diff <= 60.0*5 {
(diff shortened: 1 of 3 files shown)


rhysd:
Oh, ok, the condition was slightly incorrect. Thank you for catching it.

rhysd (approved):

PR #21, opened 2021-08-02
readme: typo

bendrucker:
Fixes a typo I spotted while while reading the docs. Really glad to have found a static validator for Actions workflows, thanks for this!


rhysd (approved):
Thanks for catching this.

PR #34, opened 2021-08-18
Add Docker image for pre-commit

rally-carlos:
Is making actionlint available as a Docker image a possibility?

This PR illustrates what the implementation could look like. An account on a Docker image repository is required.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -95,3 +95,20 @@ jobs:
       - run: make lint
       - name: Lint bash scripts
         run: shellcheck ./scripts/download-actionlint.bash ./playground/post-install.bash ./playground/deploy.bash
+  docker:
+    name: Test Dockerfile
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build image
+        id: image
+        uses: docker/build-push-action@v2
+        with:
+          build-args: |
+            GOLANG_VER=1.16
+          push: false
+      - name: Test Docker image
+        run: docker container run
+          --mount type=bind,source="$(pwd)",target=/mnt/app
+          --workdir /mnt/app
+          -- ${{ steps.image.outputs.digest }} --verbose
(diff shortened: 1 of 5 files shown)


rhysd:
I think using single binary would be better than using Docker for the following reasons:

- Overhead is larger. Running container costs more than running a single executable. The Docker image includes entire Go toolchain so it would also require much storage
- `shellcheck` and `pyflakes` integrations are not available. To enable them, we also need to install them in the image. So maintaining the image is a bit harder

I think, it is easiest way to download executable by using [the donwload script](https://github.com/rhysd/actionlint/blob/main/docs/install.md#download-script) and executing the downloaded binary directly.

rhysd:
I'm not sure since I'm not familiar with pre-commit, but you may be able to make a small `actionlint` command wrapper script and to use it from the tool.

1. Check if `actionlint` command is installed or not. If it is installed globally, the script runs it
2. Check if `actionlint` executable exists in the current directory (or somewhere like `.pre-commit` directory in your repository root). If it doesn't, download an executable by using `curl` and the download script locally.
3. The script runs the locally downloaded executable

rhysd:
> The Docker image includes entire Go toolchain so it would also require much storage

Ah, this'd be false thanks to [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds).

rhysd:
OK, I'd consider to prepare some Docker image for actionlint. I think it is useful for users don't want to install dependencies like `shellcheck` or `pyflake`.

xsc27:
I have pushed a Dockerfile to include the dependencies with Alpine Linux as the base image. To make the image smaller, may not be worth the ROI, IMO. 

I have also included a test to verify the container can build and run. Does this cover what you are looking for?

rhysd:
Thank you for the update. I think `pyflakes` is also necessary to check scripts with `shell: python`. So I think we need to build a image based on `python:3.9-alpine` and install it via `pip3 install pyflakes` as [action-actionlint does](https://github.com/reviewdog/action-actionlint/blob/ba191e76b796ad82e368815d7c54a56c5749e354/Dockerfile#L3-L5).

rhysd:
Thank you for the update. Looks good. Would you wait for a few days? I'm now improving type checking in other branch. After finishing the task, let me try this branch.

rhysd:
I confirmed it worked. I released the first version manually: https://hub.docker.com/repository/docker/rhysd/actionlint

I'll add some commits for follow up.

rhysd (approved):

rhysd:
Merged :+1: Thank you for the additions.

PR #38, opened 2021-08-23
fix(pre-commit): rename .pre-commit-hook.yaml to .pre-commit-hooks.yaml

xsc27:
🤦


rhysd (approved):
ah, thanks for catching this.

PR #43, opened 2021-09-01
Modify dockerhub url in usage.md

radomirbosak:
The original url

https://hub.docker.com/repository/docker/rhysd/actionlint

required user to log in, while the modified one

https://hub.docker.com/r/rhysd/actionlint

does not.
This change makes using docker images more accessible for users without docker account.


rhysd (approved):
Ah, thanks!

PR #52, opened 2021-09-23
Fix shell name validation example and test

fornwall:
The [shell name validation check](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-shell-names) documentation has the following example:

```yaml
  linux:
    runs-on: ubuntu-latest
    steps:
      - run: echo 'hello'
        # ERROR: Unavailable shell
        shell: dash
      - run: echo 'hello'
        # ERROR: 'powershell' is only available on Windows
        shell: powershell
      - run: echo 'hello'
        # OK: 'powershell' is only available on Windows
        shell: powershell
```

As done in this PR, the `OK: 'powershell' is only available on Windows` section should be moved to the Windows section.

### DIFF
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -1508,9 +1508,6 @@ jobs:
       - run: echo 'hello'
         # ERROR: 'powershell' is only available on Windows
         shell: powershell
-      - run: echo 'hello'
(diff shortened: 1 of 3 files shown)


rhysd (approved):
Ah, thanks for catching this!

PR #67, opened 2021-11-03
`name` is a valid property of `runner`

ioanrogers:
Not sure when it showed up on github.com, but `runner.name` has been available since at least GHE 2.22, the first GHE release with Actions.
https://docs.github.com/en/enterprise-server@2.22/actions/learn-github-actions/contexts#runner-context

I wasn't sure where to put a test, so hope my choice is okay.
 
Thanks!


rhysd:
Thank you for catching this!

The property was missing in actionlint because `runner.name` was added to the documentation at Sep. 22.

https://github.com/github/docs/pull/10230

rhysd (approved):

rhysd:
This fix was released as v1.6.7

PR #93, opened 2021-12-18
brew: remove tap setup in favor of homebrew-core one

chenrui333:
As homebrew-core [has accepted the formula](https://github.com/Homebrew/homebrew-core/pull/85361), it would be good to remove the tap in favor of the core one.

Recent analytics metrics as below:
```
$ brew info actionlint
actionlint: stable 1.6.8 (bottled)
Static checker for GitHub Actions workflow files
https://rhysd.github.io/actionlint/
/opt/homebrew/Cellar/actionlint/1.6.8 (7 files, 4.2MB) *
  Poured from bottle on 2021-12-18 at 15:41:14
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/actionlint.rb
License: MIT
==> Dependencies
Build: go ✔, ronn ✘
==> Analytics
install: 169 (30 days), 540 (90 days), 555 (365 days)
install-on-request: 169 (30 days), 540 (90 days), 555 (365 days)
build-error: 0 (30 days)
```


rhysd:
Thank you for the suggestion. But I think there is no active motivation to delete the description. I wrote instructions to install Homebrew's official package and our own package. And I put the official one at first. So almost all users would install the official one.

Our own Homebrew package is provided to ensure to take control for distributing `actionlint` command to macOS users via Homebrew (for example, Homebrew may stop providing `actionlint` package for some reasons in the future.)

PR #101, opened 2022-01-01
introduce the `runner.arch` context value

shogo82148:
I found that a new context value `runner.arch` is available.

ref. https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context

the `runner.arch` context value


rhysd (approved):
Thank you!

PR #107, opened 2022-02-01
Support reusable workflows in the same repository

jsok:
Recently announced that support for referencing workflow in the same
repository at the checked out commit:
https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

PR #115, opened 2022-02-19
fix: allow local paths for reusable workflows

### LINKED ISSUE #114: Reusable workflows referenced by path (which is valid) are not being allowed
At https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow it is stated:

> You reference reusable workflow files using one of the following syntaxes:
> 
> * {owner}/{repo}/{path}/{filename}@{ref} for reusable workflows in public repositories.
> * ./{path}/{filename} for reusable workflows in the same repository.

I have just created a reusable workflow in the same repository as the caller workflow (so matching the second bullet point from above) and am calling it like so:

```yaml
dp_deploy_dev:
    uses: ./.github/workflows/deploy-asset.yml
```

actionlint is returning this error:

```
.github/workflows/ci.yml:378:11: reusable workflow call "./.github/workflows/deploy-asset.yml" at "uses" is not following the format "owner/repo/path/to/workflow.yml@ref". see https://docs.github.com/en/actions/learn-github-actions/reusing-workflows for more details [workflow-call]
    |
378 |     uses: ./.github/workflows/deploy-asset.yml
    |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
------
```

Given the documentation quoted above I'm pretty certain this is valid syntax and actionlint is raising a false positive. Would you agree?

stabai:
fixes #114

Currently, the linter throws an error for calls to a local workflow file. However, local workflow files are perfectly valid, and the pattern is encouraged by GitHub documentation:

> You reference reusable workflow files using one of the following syntaxes:
> - {owner}/{repo}/{path}/{filename}@{ref} for reusable workflows in public repositories.
> - **/{path}/{filename} for reusable workflows in the same repository.**

https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

### DIFF
--- a/linter_test.go
+++ b/linter_test.go
@@ -160,6 +160,7 @@ func TestLinterLintError(t *testing.T) {
 				}
 
 				if len(errs) != len(expected) {
+					fmt.Printf("File: %s", base)
 					t.Fatalf("%d errors are expected but actually got %d errors: %# v", len(expected), len(errs), errs)
 				}
 
--- a/linter_test.go
+++ b/linter_test.go
@@ -160,7 +160,6 @@ func TestLinterLintError(t *testing.T) {
 				}
 
 				if len(errs) != len(expected) {
-					fmt.Printf("File: %s", base)
 					t.Fatalf("%d errors are expected but actually got %d errors: %# v", len(expected), len(errs), errs)
 				}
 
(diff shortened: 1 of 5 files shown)


rhysd:
Is this duplicate of #107?

stabai:
Yep, I wrote this before I saw 107. I'll revert now that it's been merged.

Issue #88, opened 2021-12-10
actionlint doesn't show the line the error is on in a workflow


toast-gear:
When you run actionlint in a workflow it is unable to indicate which line the error is on:

https://github.com/toast-gear/actionlint-test/pull/3/checks

<img width="996" alt="image" src="https://user-images.githubusercontent.com/15716903/145646140-2c4c49bd-2992-4040-b83a-f80b1f87a7b2.png">

rhysd:
It's not possible to point the position within multi-line string (after `|`) because go-yaml removes leading indentations and there is no way to know the length of the indentations from parsed YAML node.

jbergstroem:
Perhaps feedback can cover the full run segment instead of only `run:`?

rhysd:
What does feedback mean?

Anyway, I don't think this can be implemented otherwise go-yaml provides length of the indentation in a YAML AST node. If someone can submit a PR to implement it, I would happily review it.

jbergstroem:
> What does feedback mean?

Sorry, explained myself poorly. Feedback as in the underline and "full segments" would be all content in the run directive. 

> Anyway, I don't think this can be implemented unless go-yaml provides length of the indentation in a YAML AST node. 

I came to the same conclusion.

rhysd:
So you mean we can get the position information from the parent node of the script string node, right? I think it would be hard.

- [go-yaml's `Node` API](https://pkg.go.dev/gopkg.in/yaml.v3?utm_source=godoc#Node) does not provide the end position of the node. It only provides the start position.
- Even if we can know the end position of the parent node, we still need to parse its contents to get the indent of the child string node. I think it means we need to re-implement YAML parser partially. YAML has complex syntax so it would not be easy.

MattMencel:
I don't know if this is possible, but if it can parse the shellcheck error, you can identify which line to end at.

In this case, `SC2086:info:2:27`, it's inside the `run:` line 2 column 27. I would want the review to show through line 20 then.

![Screenshot 2024-09-18 at 9 38 32 AM](https://github.com/user-attachments/assets/46dacd7d-7f6c-4d15-b0ad-63be3003c671)

rhysd:
That strategy doesn't work in some cases. for example

```yaml
- run: "echo foo\necho bar"
```

However it might be a better workaround than the current implementation (I'm not sure).

PR #118, opened 2022-02-25
Add configurable rule to avoid overlong inline `run:` scripts

cmur2:
Hi, this is my first contribution here since I wanted to learn how to add new rules. I'm prepared to iterate on the idea (which is opinionated) if maintainers like it.

Motivation: sometimes I see / author GHA workflows that have long bash scripts (more than a dozen lines) in `run:` steps. This makes the overall workflow harder to read.
It also makes `shellcheck` warnings from reported by actionlint more difficult to interpret due to the "nested" line numbers - I first need to figure out which `run:` step it is about and then which line inside that step `shellcheck` complains. This gets a bit easier when the scripts are shorter.

Usually short scripts can be achieved by refactoring into separate `.sh` files or build systems (`Makefile` etc.). This is why I'm proposing the idea that actionlint could detect those situations.

I also added a configuration parameter since the amount of lines where to make the cut is probably subjective.

Please tell me what you think! No hard feelings if you think it is too opinionated 😄 

PS: [build](https://github.com/cmur2/actionlint/actions/runs/1898186503) in my fork is green.

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,6 +16,7 @@ type Config struct {
 		// Labels is label names for self-hosted runner.
 		Labels []string `yaml:"labels"`
 	} `yaml:"self-hosted-runner"`
+	RunScriptMaxLines *int `yaml:"run-script-max-lines",omitempty`
 }
 
 func parseConfig(b []byte, path string) (*Config, error) {
@@ -39,6 +40,8 @@ func writeDefaultConfigFile(path string) error {
 	b := []byte(`self-hosted-runner:
   # Labels of self-hosted runner in array of string
   labels: []
+
+#run-script-max-lines: 23
 `)
 	if err := ioutil.WriteFile(path, b, 0644); err != nil {
 		return fmt.Errorf("could not write default configuration file at %q: %w", path, err)
--- a/linter.go
+++ b/linter.go
@@ -457,8 +457,12 @@ func (l *Linter) check(path string, content []byte, project *Project, proc *conc
 		dbg := l.debugWriter()
 
 		var labels []string
+		var maxLines int = 10
 		if cfg != nil {
 			labels = cfg.SelfHostedRunner.Labels
+			if cfg.RunScriptMaxLines != nil {
+				maxLines = *cfg.RunScriptMaxLines
+			}
 		}
 
 		rules := []Rule{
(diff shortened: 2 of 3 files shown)


rhysd:
> sometimes I see / author GHA workflows that have long bash scripts (more than a dozen lines) in run: steps. This makes the overall workflow harder to read.

I think this motivation is more general than checking GitHub Actions workflow. Since too long line is not good for maintenance not only for scripts, but also for every line of configuration, I think checking the line length generally by some YAML linters would be better rather than checking this by actionlint. For example, [`line-length`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.line_length) option of `yamllint` command.

cmur2:
An example of what I mean would be this:

```yaml
---
name: example

on: push

jobs:
  example:
    runs-on: ubuntu-latest
    steps:
    - name: Do things
      run: |
        # line 1 of a e.g. Bash script that could also be in a file
        # line 2
        # line 3
        # line 4
        # line 5
        # line 6
        # line 7
        # line 8
        # line 9
        # line 10
        # line 11
        # line 12 and so on
```

A real-life example would be from my own repository: https://github.com/cmur2/dyndnsd/blob/dbc61d72fbdf3b02b28452221ee43d2cf8d59d2d/.github/workflows/vulnscan.yml#L29-L46 <-- this kind of script makes the workflow harder to read and shellcheck warnings harder to interpret

toast-gear:
> It also makes shellcheck warnings from reported by actionlint more difficult to interpret due to the "nested" line numbers - I first need to figure out which run: step it is about and then which line inside that step shellcheck complains

https://github.com/rhysd/actionlint/issues/88 thought I'd just mention I'd very much appreciate something that makes it easier to see which line shellcheck is actually complaining about when it comes into multiline inline shell scripts (I'm not personally fussed about the ability to enforce a max length either number of lines or length). Thought I'd just raise that this is something that at least 1 consumer of this excellent project would like.

cmur2:
I'm also not too attached to the idea of this PR - I wanted to get familiar with the code of actionlint + do something useful.

If it does not gain traction with the maintainer, I'll close it eventually.

PR #131, opened 2022-04-11
fix: add permissions "pages" and "discussions"

suzuki-shunsuke:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions


rhysd:
Nice catch! Thanks for updating the list.

suzuki-shunsuke:
Oh, CI failed. I'm checking.
https://github.com/rhysd/actionlint/runs/5983766636?check_suite_focus=true

PR #134, opened 2022-04-26
macos-12 is now available

shogo82148:
Public beta of macOS 12 for GitHub-hosted runners is now available.

https://github.blog/changelog/2022-04-25-github-actions-public-beta-of-macos-12-for-github-hosted-runners-is-now-available/

And the document is also updated.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

image


rhysd (approved):
Thank you! Quality of changes are really good.

PR #135, opened 2022-04-26
bump actions that support node16

shogo82148:
Currently, some actions run on the node12 runtime.
But, Node.js 12 ends its support in 30 Apr 2022.
I bumped these actions.


rhysd (approved):
Thank you for many updates not only in workflows, but also in documents

PR #154, opened 2022-06-20
Add lint workflow_call input conflict with required and default

sksat:
I want to lint below invalid reusable workflow

```
on:
  workflow_call:
    inputs:
      test_input:
        description: test
        type: string
        required: true  # set this input is required
        default: a      # but default value is provided
```

When we call this workflow without `test_input`, it results like this. I made this mistake sometimes.
```
The workflow is not valid. Input test_input is required, but not provided while calling.
```


rhysd (approved):

rhysd:
This change was included in the latest release [v1.6.14](https://github.com/rhysd/actionlint/releases/tag/v1.6.14).

sksat:
Sorry, I forgot to add tests... I'll be careful next time. Thanks for release!

PR #157, opened 2022-06-23
add stdin-filename support

arahatashun:
## What
add support for overwriting file names when using stdin.
## Why
When I tried developing a VSCode extension for actionlint, I felt it would be helpful if we had this option.
FYI: https://github.com/fnando/vscode-linter/blob/main/docs/creating-linters.md

For example, eslint has this kind of option. https://eslint.org/docs/user-guide/command-line-interface#--stdin-filename
## QA
With --stdin-filename 
```
shun@Mac-mini actionlint % echo "foo"|  ./actionlint --stdin-filename "test.yml" -
test.yml:1:1: workflow is scalar node but mapping node is expected [syntax-check]
  |
1 | foo
  | ^~~
test.yml:1:1: "jobs" section is missing in workflow [syntax-check]
  |
1 | foo
  | ^~~
```
Without --stdin-filename 
```
shun@Mac-mini actionlint % echo "foo"|  ./actionlint -
<stdin>:1:1: workflow is scalar node but mapping node is expected [syntax-check]
  |
1 | foo
  | ^~~
<stdin>:1:1: "jobs" section is missing in workflow [syntax-check]
  |
1 | foo
  | ^~~
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -98,8 +98,8 @@ func (cmd *Command) runLinter(args []string, opts *LinterOptions, initConfig boo
 			return nil, fmt.Errorf("could not read stdin: %w", err)
 		}
 		n := "<stdin>"
-		if opts.StdInFileName != "" {
-			n = opts.StdInFileName
+		if opts.StdinFileName != "" {
+			n = opts.StdinFileName
 		}
 		return l.Lint(n, b, nil)
 	}
@@ -142,7 +142,7 @@ func (cmd *Command) Main(args []string) int {
 	flags.BoolVar(&opts.Verbose, "verbose", false, "Enable verbose output")
 	flags.BoolVar(&opts.Debug, "debug", false, "Enable debug output (for development)")
 	flags.BoolVar(&ver, "version", false, "Show version and how this binary was installed")
-	flags.StringVar(&opts.StdInFileName, "stdin-filename", "", "File name for stdin")
+	flags.StringVar(&opts.StdinFileName, "stdin-filename", "", "File name for stdin")
 	flags.Usage = func() {
 		fmt.Fprintln(cmd.Stderr, commandUsageHeader)
 		flags.PrintDefaults()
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -97,7 +97,11 @@ func (cmd *Command) runLinter(args []string, opts *LinterOptions, initConfig boo
 		if err != nil {
 			return nil, fmt.Errorf("could not read stdin: %w", err)
 		}
-		return l.Lint("<stdin>", b, nil)
+		if opts.StdInFileName == "" {
+			return l.Lint("<stdin>", b, nil)
+		} else {
+			return l.Lint(opts.StdInFileName, b, nil)
+		}
```
rhysd, line 104:
```suggestion
		n := "<stdin>"
		if opts.StdInFileName != "" {
			n = opts.StdInFileName
		}
		return l.Lint(n, b, nil)
```

Would you reduce repeating code?

linter.go
```diff
@@ -79,6 +79,8 @@ type LinterOptions struct {
 	// Format is a custom template to format error messages. It must follow Go Template format and
 	// contain at least one {{ }} placeholder. https://pkg.go.dev/text/template
 	Format string
+	// Set FileName for stdin
+	StdInFileName string
```
rhysd, line 83: Could you rename this to `StdinFileName` since stdin is one word?

rhysd (changes requested):

rhysd:
@arahatashun When you finish to build the VS Code plugin, I would be glad to hear that. I'll update document to include the URL to the project.

rhysd (approved):

rhysd:
Thanks!

rhysd:
This change was included in the latest release [v1.6.14](https://github.com/rhysd/actionlint/releases/tag/v1.6.14).

arahatashun:
Thank you for the quick release. I published the VSCode extension <https://marketplace.visualstudio.com/items?itemName=arahata.linter-actionlint>.

itsmeboy0:
@rhysd Can you add a reference to it in the docs now?

rhysd:
I'm sorry for catching the comment lately since this issue was already closed. I added description about the VS Code extension. Thank you for making it!

https://github.com/rhysd/actionlint/commit/49dede4f36dcb71b90d1847094458a5562295e9b

PR #159, opened 2022-06-27
Add multi platform build for Docker image

politician:
This adds support for Mac M1 (and other arm processors)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -50,14 +50,20 @@ jobs:
         id: tag
         run: |
           echo "::set-output name=name::${GITHUB_REF#refs/tags/v}"
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
       - name: Login to DockerHub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         with:
           username: rhysd
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build and push
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v3
         with:
+          platforms: linux/amd64,linux/arm64
           build-args: |
             GOLANG_VER=1.18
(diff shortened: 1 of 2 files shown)


rhysd:
@politician Thank you for adding this. How did you confirm this change worked?

politician:
@rhysd Sorry it wasn't working due to Shellcheck arm64 not being present in the Alpine repo (?). I have modified the Dockerfile and repushed.

- [Docker Workflow running successfully on my repo](https://github.com/politician/actionlint/runs/7066670877?check_suite_focus=true) 
- [Image built on my Docker hub](https://hub.docker.com/r/iromain/actionlint/tags)
- Running it on my Mac M1:

Screenshot 2022-06-27 at 12 29 20

rhysd:
@politician Thank you for checking that. From your screenshot, it fails with YAML parse error. In the case, shellcheck is not run. Could you confirm the copied `shellcheck` binary works fine in the built image? Since only `/usr/local/bin/shellcheck` is copied, it may not work when it loads some dynamic link libraries at runtime.

politician:
@rhysd in the ShellCheck Dockerfile, they do the same [copy operation](https://github.com/koalaman/shellcheck/blob/f1bdda54cb87918f30bf11afa94dad0de149c324/Dockerfile.multi-arch#L25)

ShellCheck works:
Screenshot 2022-06-27 at 13 33 09

rhysd:
> in the ShellCheck Dockerfile, they do the same [copy operation](https://github.com/koalaman/shellcheck/blob/f1bdda54cb87918f30bf11afa94dad0de149c324/Dockerfile.multi-arch#L25)

Oh, ok, then it should work (with all dependencies being statically linked?). Thanks for the confirmation

rhysd (approved):

rhysd:
@politician Thank you! Merged.

rhysd:
actionlint v1.6.15 was released and now `linux/arm64` container is available.

https://hub.docker.com/layers/rhysd/actionlint/1.6.15/images/sha256-f63ee59f1846abce86ca9de1d41a1fc22bc7148d14b788cb455a9594d83e73f7?context=repo

PR #168, opened 2022-07-06
Fix scoop installation link

sitiom:
actionlint has been moved in to the [main repo](https://scoop.sh/#/apps?q=actionlint&s=0&d=1&o=true)


rhysd (approved):

rhysd:
Thanks for catching that. Linking to the search page seems better.

PR #190, opened 2022-08-08
fix: fix validation of github.triggering_actor

stefreak:
`github` not just exposes `actor`, but also `triggering_actor`.


rhysd (approved):
Thanks!

PR #199, opened 2022-08-23
remove strategy from disallowed things in reusable workflows

### LINKED ISSUE #197: When a reusable workflow is called with "uses", "strategy" is not available.
I'm getting the following action lint error when I am using a reusable workflow with a matrix strategy.

`.github/workflows/lint-and-format.yaml:12:5: when a reusable workflow is called with "uses", "strategy" is not available. only following keys are allowed: "name", "uses", "with", "secrets", "needs", "if", and "permissions" in job "python-pr-lint" [syntax-check]`

As per the documentation, this is valid, and the workflow actually runs fine. Can action lint please be updated to match the documentation? - https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-matrix-strategy-with-a-reusable-workflow

seemethere:
Resolves #197

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -1204,7 +1204,6 @@ func (p *parser) parseJob(id *String, n *yaml.Node) *Job {
 				"if",
 				"steps",
 				"timeout-minutes",
-				"strategy",
 				"continue-on-error",
 				"container",
 				"services",


rhysd:
This does not solve #197. The line is only used for making an error message. Please check your fix before making a PR.

Issue #198, opened 2022-08-23
Feature request: add option to enforce security best practices


stefreak:
It would be nice to add an option to check for security best practices.

Best practices that could be supported:
- [set minimum `GITHUB_TOKEN` permissions](https://github.com/step-security/secure-workflows#1-automatically-set-minimum-github_token-permissions)
- [Pin Actions to a full length commit SHA](https://github.com/step-security/secure-workflows#2-pin-actions-to-a-full-length-commit-sha)

Some other tools can already check for these things, like [OSSF scorecard](https://github.com/ossf/scorecard) and [StepSecurity SecureWorkflow](https://github.com/step-security/secure-workflows).

Actionlint has an advantage over these tools though: It really understands advanced features like custom actions, reusable workflows etc. The tools I named are also hard to set up for private repositories, and actionlint is really easy to set up for private repositories.

ansraliant:
Actually I was just thinking the same haha. I am currently trying to implement something.

I wanted to start (at least for now) on checking that we have the permissions set to none (or {} in gh actions jargon) at the beginning of the job.

wolfeidau:
@stefreak I build a small program using actionlint as a library to solve the [Pin Actions to a full length commit SHA](https://github.com/step-security/secure-workflows#2-pin-actions-to-a-full-length-commit-sha) problem, turns out there is a lot of "new" features needed to interact with the GitHub API. https://github.com/wolfeidau/github-action-workflow-check 

Rate limits are so low you need to authenticate, which in turn requires other security features.

PR #209, opened 2022-09-06
Enforce empty permissions

ansraliant:
## Description 

Following the [Security Hardening for Github Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) guide, they mention the use of permissions. 

Permissions should be given as needed, and have them write-all should be avoided. 

One way to accomplish this recommendation is to set the global permissions to none, and give on demand to the jobs that need it.

## Implementation details

Enforcing such approach perhaps is not desired for everyone, so I added a configuration option. The default values is `false`, which keeps the current behaviour, but if the user sets it to `true`, then, the linter will check the top level for empty permissions.

## Considerations

I added a example yaml, which is correct workflow file, but that's where the test end. Please let me know if you want more tests, for example different variations of the permissions array, I'll gladly add them.

## Related Issues

Tries to solve, at least partially, one of the aspects discussed in #198

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -624,9 +624,16 @@ func (p *parser) parseEvents(pos *Pos, n *yaml.Node) []Event {
 }
 
 // https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions
-func (p *parser) parsePermissions(pos *Pos, n *yaml.Node) *Permissions {
+func (p *parser) parsePermissions(pos *Pos, n *yaml.Node, config *Config) *Permissions {
 	ret := &Permissions{Pos: pos}
 
+	if config != nil && config.EnforceEmptyPermissions && pos.IsTopLevel() {
+		if !isNull(n) {
+			p.errorf(n, "expected empty Permissions. If this is a mistake, please disable "+
+				"enforce_empty_permissions in the config file.")
+		}
+		return ret
+	}
 	if n.Kind == yaml.ScalarNode {
 		ret.All = p.parseString(n, false)
 	} else {
@@ -640,7 +647,6 @@ func (p *parser) parsePermissions(pos *Pos, n *yaml.Node) *Permissions {
 		}
 		ret.Scopes = scopes
 	}
-
 	return ret
 }
 
@@ -1070,7 +1076,7 @@ func (p *parser) parseSteps(n *yaml.Node) []*Step {
 }
 
 // https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_id
-func (p *parser) parseJob(id *String, n *yaml.Node) *Job {
+func (p *parser) parseJob(id *String, n *yaml.Node, config *Config) *Job {
 	ret := &Job{ID: id, Pos: id.Pos}
 	call := &WorkflowCall{}
 
@@ -1112,7 +1118,7 @@ func (p *parser) parseJob(id *String, n *yaml.Node) *Job {
 			}
 			stepsOnlyKey = k
 		case "permissions":
-			ret.Permissions = p.parsePermissions(k.Pos, v)
+			ret.Permissions = p.parsePermissions(k.Pos, v, config)
 		case "environment":
 			ret.Environment = p.parseEnvironment(k.Pos, v)
 			stepsOnlyKey = k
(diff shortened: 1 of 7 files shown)


ansraliant:
@rhysd can't assign reviewers tho haha. A quick look at previous PRs seems you are the default reviewer for PRs :)

parse.go
```diff
@@ -624,9 +624,16 @@ func (p *parser) parseEvents(pos *Pos, n *yaml.Node) []Event {
...
 
 // https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions
-func (p *parser) parsePermissions(pos *Pos, n *yaml.Node) *Permissions {
+func (p *parser) parsePermissions(pos *Pos, n *yaml.Node, config *Config) *Permissions {
 	ret := &Permissions{Pos: pos}
 
+	if config != nil && config.EnforceEmptyPermissions && pos.IsTopLevel() {
+		if !isNull(n) {
+			p.errorf(n, "expected empty Permissions. If this is a mistake, please disable "+
+				"enforce_empty_permissions in the config file.")
```
stefreak, line 633: Some teams use `read-all` as the default, or `contents`, instead of empty.

What do you think about an option to make sure that
- default permissions are reduced in the beginning of a workflow
- default permissions are read-only

On the other hand, maybe it's also a good idea to go with one standard instead of having multiple ways in the community.

What are your thoughts on this @rhysd?

stefreak:
Loving this change, it will make a lot of github workflows more secure 👍

parse.go
giner: I like the idea of enforcing no permissions. This would raise security awareness in areas of GitHub Actions which are not obvious.

ansraliant: Without setting the var in the config the behaviour is the same as we currently have, which I believe is the standard, as GitHub docs don't force setting any particular permissions.
This option is mostly concerned for those who want to follow the security hardening guidelines. 

> Some teams use read-all as the default, or contents, instead of empty.

Regarding this aspect, perhaps I can tune the config, and add a couple of options. For example, instead of just `enforce_empty_permissions`, something like `enforce_permissions` with 2 options read-only and empty instead of the current boolean.

> On the other hand, maybe it's also a good idea to go with one standard instead of having multiple ways in the community.

This one might require more discussion with @rhysd. I believe we should have a standard, but also allow the user to configure the linter to their needs. 
Should we go with the most secure options right of the bat and have the users disable those options which are too secure?

ansraliant:
@rhysd ping

Do you think we should set by default the most strict permissions and let the user then, make it more permissive by config file?
Also, if this approach to add a new option looks viable to you, I also am interested in adding more once we decide on an approach.

rhysd:
I agree that security of workflow should be considered and enforcing the security is good idea.

However I don't think this check is currently useful because setting and maintaining proper minimal permission scopes is too hard.
To know proper permissions, users need to check what permissions are necessary for each actions in each steps. What permissions are used by action is usually not described in the action's document. So users need to check its implementation (usually what `git` commands are used and what methods of `octokit` client are called). I'm afraid that users eventually set `permissions: write-all` to avoid the error.

And I don't think introducing new config only for this check since no one wants to maintain configuration just for checking CI workflows.

Currently actionlint only has rules which are enabled by default. But I'm thinking how to handle optional rules, which would be useful for everyone but may cause some false-positives in some cases.

giner:
> However I don't think this check is currently useful because setting and maintaining proper minimal permission scopes is too hard. To know proper permissions, users need to check what permissions are necessary for each actions in each steps. What permissions are used by action is usually not described in the action's document.

While I understand the concern as someone who is doing permissions hardening for a while now I wouldn't say it's too hard. In most cases it is just "contents: read". Either way whether it is the default or not such an option would be very useful.

giner:
> I'm afraid that users eventually set permissions: write-all to avoid the error.

One more point on this, with the suggested enforcement "permissions" settings become explicit and reviewable, i.e. workflow creators become aware of them, which is on the plus side.

ansraliant:
What if, instead of adding a configuration option, we add a separate mode to the linter. Something like `actionlint --enforce-permissions`?

That way it we would:
 * not add extra config options thus keeping the config file as is
 * only advanced users that know how the github token permissions work enable it, thus not affecting in any way the other users

I can rework this PR, and perhaps we can see how that goes, what do you think @rhysd

Issue #234, opened 2022-10-12
set-output is deprecated


larrys:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

I am getting the following warning when running actionlint

The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Gudahtt:
actionlint itself is using the deprecated command. e.g. https://github.com/rhysd/actionlint/blob/main/scripts/download-actionlint.bash#L135

Catching usage of this deprecated command would be useful, but I don't think that's what this issue was requesting.

rhysd:
> you mean statically check if a command is deprecated?

No. I meant statically detecting the command is used.

> actionlint itself is using the deprecated command.

Yes. I said this idea would be great if it can be realized without false positive in https://github.com/rhysd/actionlint/issues/234#issuecomment-1277549903.

rhysd:
Today I have some time to implement this with simple pattern matching. I will try it and look if it causes some false positives or not.

rhysd:
[v1.6.22](https://github.com/rhysd/actionlint/releases/tag/v1.6.22) included this feature.

PR #239, opened 2022-10-25
Replace deprecated `::set-output` command

Mrtenz:
As mentioned in #234, the `::set-output` is deprecated. This would cause warnings to show up when using the installation script. I've replaced it in that script, as well as all the other usages of `::set-output` I could find.


rhysd:
This patch was merged into main branch by cherry-picking at https://github.com/rhysd/actionlint/commit/453276a8004cc29462fbbafd61fc4f3754af3f9a. Closing.

Mrtenz:
Thanks!

PR #257, opened 2023-01-15
Make Actionlint Directly Invokable as an Action Itself 

norwd:
This adds an `action.yml` file that downloads and installs the `actionlint` executable and runs it with options specified in the workflow file. Since the `scripts/download-actionlint.bash` file can get run in an action, the file also now uses the workflow commands to report errors/warnings etc... during the download.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -18,7 +18,7 @@ runs:
     - name: "Report Invalid Color Input"
       if: inputs.color != 'yes' && inputs.color != 'true' && inputs.color != 'no' && inputs.color != 'false'
       shell: bash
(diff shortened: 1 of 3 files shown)


aschwenn:
would love to see this usable in github cis in this way

rhysd:
This should be done in separate repository because tags of this repository is for `actionlint` command. In addition, reusing the download script for executing actionlint is confusing and a bit hard to maintain.

Please make an issue and discuss there at first. Honestly I'm not seeing much difference between running the download script then executing the downloaded binary with `run:` and running the dedicated action with `uses:`.

norwd:
> Please make an issue and discuss there first.

Will do 👍, please see #262

PR #269, opened 2023-02-03
Add docs for using actionlint via trunk

dapirian:
[trunk](https://docs.trunk.io/docs) is an extendable superlinter with a builtin language server and preexisting issue detection. Actionlint is integrated [here](https://github.com/trunk-io/plugins).

We're big fans of the actionlint project and hope our integration can increase its popularity!


rhysd (approved):
Thanks for integrating to your service.

PR #270, opened 2023-02-03
Support disallowing specific runner labels

andersfischernielsen:
This extends the configuration options with a new `disallowed-runners` property containing `labels` that disallow certain `runs-on` labels. The disallowed labels can include known runner labels for scenarios where certain standard runners aren't desirable.

This is useful e.g. when migrating from one runner label to another in shared workflows to inform authors of changes made to workflows that certain labels are no longer allowed.

For example, this allows migrating away from e.g. `ubuntu-20.04` to `ubuntu-22.04` by configuring 

``` yaml
# File: actionlint.yaml

disallowed-runners: 
  labels: 
    - ubuntu-20.04
```

to mark the `ubuntu-20.04` label as not allowed.

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,6 +16,11 @@ type Config struct {
 		// Labels is label names for self-hosted runner.
 		Labels []string `yaml:"labels"`
 	} `yaml:"self-hosted-runner"`
+	// DisallowedLabels is configuration for disallowed runner labels.
+	DisallowedRunner struct {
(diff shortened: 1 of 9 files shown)


andersfischernielsen:
@rhysd what's your take on this? 😄

andersfischernielsen:
@rhysd Should I just close this, or? 👀

PR #282, opened 2023-03-11
Add workflow usage example

rajbos:
I keep searching for the example here and then need to back to one of my other repos that uses it 😄. I think it will be helpful for others as well.


rajbos:
I see this is documented [here](https://github.com/rhysd/actionlint/blob/main/docs/usage.md), so no need to include it in the README. Closing the PR.

PR #268, opened 2023-02-02
docs: example of how to pre-install actionlint when using pre-commit

gilesw:

PR #287, opened 2023-03-23
Install actionlint in pre-commit

killermoehre:
`pre-commit` can actually install all the necessary tools, if the proper language tag is configured.

This decouples the need to install `actionlint` separately.


s-weigand:
I think that won't work [see this comment](https://github.com/rhysd/actionlint/pull/268#issuecomment-1426318829)

killermoehre:
I see. Bummer. But the workaround with `additional_dependencies:` works, so I'm fine.

PR #289, opened 2023-03-27
Remove outdated Problem Matchers `-color` note

carlcsaposs-canonical:
It looks like support for `-color` option with Problem Matchers was added in 62510ae8f7d17fd5313d76a13436f1de255fe2d8


rhysd (approved):
Thanks for catching this.

PR #290, opened 2023-03-30
Stronger external checks

arnodb:
Hi, I have just started using `actionlint` and it is a great tool indeed.

I was just surprised that shell checks were disabled silently on my machine because the command was not found.

This pull request aims at solving this small unsafety. The consequence is that some invocations may fail and the user shall disable (or fix) shellcheck/pyflakes explicitly. Default values remain unchanged.

I added tests to check the new behaviour.

### DIFF
--- a/linter.go
+++ b/linter.go
@@ -504,7 +504,8 @@ func (l *Linter) check(
 			if err == nil {
 				rules = append(rules, r)
 			} else {
-				l.log("Rule \"shellcheck\" was disabled:", err)
+				l.log("Error in rule \"shellcheck\":", err)
+				return nil, err
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


rhysd:
This is a good point. When `shellcheck` doesn't exist though `-shellcheck` option is set, it would be surprising that `actionlint` continues without enabling checks with shellcheck.

rhysd (approved):
This PR needs some small fixes, but I'll do that after merging this.

rhysd:
Unfortunately this doesn't work with the default behavior of actionlint. It tries to use `shellcheck` command as default behavior and ignore an error when the command is not found.

PR #293, opened 2023-04-15
Remove `version` parameter in Winget workflow

sitiom:
The `version` parameter is optional and derived from the `release-tag`, which usually cuts the `v` prefix from it. Otherwise, it would lead to an incorrect `PackageVersion` such as https://github.com/microsoft/winget-pkgs/pull/102052.

### DIFF
--- a/.github/workflows/winget.yaml
+++ b/.github/workflows/winget.yaml
@@ -17,5 +17,4 @@ jobs:
           identifier: rhysd.actionlint
           installers-regex: '_windows_\w+\.zip$'
           token: ${{ secrets.WINGET_TOKEN }}
-          version: ${{ inputs.tag }}
           release-tag: ${{ inputs.tag }}


rhysd (approved):

rhysd:
Thanks. I apologize for the delay.

rhysd:
@sitiom Winget workflow fails after this PR. Do you know the reason? https://github.com/rhysd/actionlint/actions/runs/5280674972/jobs/9553198795

sitiom:
Maybe it could be that your token is invalid, but I also see the mismatched installer URLs, which should be a Komac bug. @russellbanks

rhysd:
I checked the token, but it was valid and not expired. I know that the same token worked before.

image

> but I also see the mismatched installer URLs, which should be a Komac bug.

OK, I will make an issue tomorrow.

sitiom:
Oh, an issue is already open for it:
- https://github.com/russellbanks/Komac/issues/215

rhysd:
Thanks. Then I can track the ticket.

sitiom:
Meanwhile, I have manually created a PR for the new version:
- https://github.com/microsoft/winget-pkgs/pull/109924

rhysd:
Thank you. It's very helpful.

Issue #297, opened 2023-04-28
Support 'go install ./...' ?


adamchainz:
The current `.pre-commit-hooks.yaml` file allows running actionlint through pre-commit's `system` and `docker` languages. These are both ways of skipping pre-commit's packaging system.

pre-commit has a [`golang` language](https://pre-commit.com/#golang), and in recent version 3.0.0 it gained the ability to bootstrap golang. This allows users to compile and run native golang tools without any setup other than pre-commit. It would be neat if actionlint supported this style of running, since it would avoid teams needing to all install actionlint themselves, or Docker.

I just tried using actionlint under pre-commit with `golang`, just this change:

```diff
--- .pre-commit-hooks.yaml
+++ .pre-commit-hooks.yaml
@@ -2,7 +2,7 @@
 - id: actionlint
   name: Lint GitHub Actions workflow files
   description: Runs actionlint to lint GitHub Actions workflow files
-  language: system
+  language: golang
   types: ["yaml"]
   files: "^.github/workflows/"
   entry: actionlint

```

It fails though, with:

```
An unexpected error has occurred: CalledProcessError: command: ('/opt/homebrew/bin/go', 'install', './...')
return code: 1
stdout: (none)
stderr:
    go: downloading github.com/yuin/goldmark v1.5.4
    go: downloading gopkg.in/yaml.v3 v3.0.1
    go: downloading github.com/fatih/color v1.15.0
    go: downloading github.com/mattn/go-colorable v0.1.13
    go: downloading github.com/mattn/go-runewidth v0.0.14
    go: downloading github.com/robfig/cron v1.2.0
    go: downloading golang.org/x/sync v0.1.0
    go: downloading golang.org/x/sys v0.6.0
    go: downloading github.com/mattn/go-isatty v0.0.18
    go: downloading github.com/rivo/uniseg v0.4.4
    package github.com/rhysd/actionlint/playground
    	imports syscall/js: build constraints exclude all Go files in /opt/homebrew/Cellar/go/1.20.3/libexec/src/syscall/js
```

`pre-commit` runs `go install ./...`, but this fails in the actinolint repo. I'm afraid I don't know anything about Go, so that's where I stop. I imagine the repository may need a little rearrangement, or a rule added to not install the playground by default... Anyway I would just like to see this happen to make it a little bit easier to run actionlint.

Freed-Wu:
It works for me.

```yaml
# ...
  - repo: https://github.com/Freed-Wu/actionlint
[...]

PR #301, opened 2023-05-24
change language of .pre-commit-hooks.yaml to golang

Freed-Wu:
Refer <https://github.com/scop/pre-commit-shfmt/blob/main/.pre-commit-hooks.yaml>

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,10 +2,11 @@
 - id: actionlint
   name: Lint GitHub Actions workflow files
   description: Runs actionlint to lint GitHub Actions workflow files
-  language: system
+  language: golang
   types: ["yaml"]
   files: "^.github/workflows/"
   entry: actionlint
+  minimum_pre_commit_version: 3.0.0
 - id: actionlint-docker
   name: Lint GitHub Actions workflow file Docker
   description: Runs actionlint Docker image to lint GitHub Actions workflow files


dokempf:
Actually, this does not work straight away, as @adamchainz noted in #297 . I found a solution on my fork that works: https://github.com/dokempf/actionlint/commit/691f4ee697dca41cc1481ee03fc43b2d35f20f98

Freed-Wu:
What is your enviroment? This is mine:
```shell
$ go version
go version go1.20.4 linux/amd64
$ uname -r
6.3.2-arch1-1
```

Freed-Wu:
If I uninstalled go, I'll get

```
An unexpected error has occurred: CalledProcessError: command: ('go', 'install', './...')
return code: 1
expected return code: 0
stdout:
    Executable `go` not found
```

Can you install go to validate again?

rhysd:
I'm not a user of pre-commit. So I'm not familiar with the tool.

Considering compatibility, should we replace `language: system` with `language: golang`? Or is it better to add new configuration `actionlint-golang` like `actionlint-docker`? I'm not sure replacing it is ok. For example, some user may use `actionlint` by downloading it from releases page and they may not have Go environment. Is the `language: golang` fallbacks to system-installed `actionlint` binary when Go toolchain is not installed?

rhysd:
Due to #297, this PR is not mergeable. Can we close this?

rhysd:
One more point. pre-commit seems to install the binary using `go install ./...`. It installs **all** executable packages in this repository. It means, all unnecessary binaries such as [the generate scripts](https://github.com/rhysd/actionlint/tree/main/scripts) are built and installed as well on your machine. Is it OK? Does pre-commit sandbox the installation in some temporary directory? Otherwise it will pollute your `~/go/bin` with the unnecessary binaries.

dokempf:
I guess one could argue how breaking the change is. I would not consider the requirement to update pre-commit to be too breaking. The fact that your existing local installation of `actionlint` is silently ignored after the change is a bit more troublesome. If you want to apply a very conservative philosophy, introducing a new `actionlint-golang` is a good choice.

Freed-Wu:
> > I would recommend adding a pre-commit version constraint to .pre-commit-hooks.yml to enforce >=3.0.0.
>
> It makes sense.

Fixed.

```yaml
  minimum_pre_commit_version: 3.0.0
```

rhysd:
@dokempf I may be too conservative. Since I'm not a user of pre-commit, I'd like to prioritize actual users' opinions. So I think changing the `language` config is OK.

In addition, I think it is good to add the classic `language: system` config as `actionlint-system` for those who don't have Go toolchain. They can continue to use the same environment with very small effort (changing pre-commit hook ID from `actionlint` to `actionlint-system`)

@Freed-Wu @dokempf Current configuration clones the HEAD of main branch, am I correct? If so, I think it is safer to fix the revision cloned by pre-commit to the latest tag name (currently it's [v1.6.24](https://github.com/rhysd/actionlint/tree/v1.6.24)). Is it possible?

rhysd:
I understood that the revision is in `.pre-commit-config.yaml` hence it is set by users. So we don't need to set it.

https://pre-commit.com/#pre-commit-configyaml---repos

rhysd (approved):

rhysd:
[v1.6.25](https://github.com/rhysd/actionlint/releases/tag/v1.6.25) was released.

adamchainz:
I have just tried the new golang hook out on my client's project and it works well. It's 10x faster than the Docker one - taking a 0.91s runtime down to 0.09s. Thank you very much! 🤘

PR #308, opened 2023-06-16
Change Winget job runner to `ubuntu-latest`

sitiom:
Unrelated to the issue in #293, but Winget Releaser now supports non-Windows runners.


rhysd (approved):

rhysd:
Thanks. `ubuntu-latest` runner would be faster in general.

PR #322, opened 2023-07-22
Upgrade `github.com/robfig/cron` and `golang.org/x/sys`

harryzcy:
Upgrade `golang.org/x/sys`
Upgrade `github.com/robfig/cron` to v3
- The package is non breaking in this case. It's not including optional second in cron spec by default. But we are using `cron.NewParser(cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)` anyways, so the parser behavior won't be changed.
- The error output of invalid syntax will be changed in the following way:

Before:
```
test.yaml:4:13: invalid CRON format "0 */3 * *" in schedule event: Expected exactly 5 fields, found 4: 0 */3 * * [events]
```

After:
```
test.yaml:4:13: invalid CRON format "0 */3 * *" in schedule event: expected exactly 5 fields, found 4: [0 */3 * *] [events]
```


rhysd (approved):

rhysd:
Looks good. Thanks for catching this.

PR #325, opened 2023-07-28
Rename newConcurrentProcess in NewConcurrentProcess

hugo-syn:
Rename _newConcurrentProcess_ in _NewConcurrentProcess_ to be accessible from the Golang library, otherwise It's not possible to call rules such as _NewRuleShellcheck_

Thanks for this incredible work :)


rhysd (approved):
Looks good. I think `concurrentProcess` also should be public. I'll do it after merging this PR.

PR #327, opened 2023-07-30
Changes to make actionlint usable when using it as a library

hugo-syn:
I've added a constructor for _RuleBase_ and changed _error_ and _errorf_ from private to public, otherwise It's not possible to create new rules. I hope that this is the best option, if not, how should we create new rules when using _actionlint_ as a library ? 

I also changed _wait_ from private to public in _process.go_.

### DIFF
--- a/linter.go
+++ b/linter.go
@@ -342,7 +342,7 @@ func (l *Linter) LintFiles(filepaths []string, project *Project) ([]*Error, erro
 		})
 	}
 
-	proc.wait()
+	proc.Wait()
 	if err := eg.Wait(); err != nil {
 		return nil, err
 	}
@@ -401,7 +401,7 @@ func (l *Linter) LintFile(path string, project *Project) ([]*Error, error) {
 	localActions := NewLocalActionsCache(project, dbg)
 	localReusableWorkflows := NewLocalReusableWorkflowCache(project, l.cwd, dbg)
 	errs, err := l.check(path, src, project, proc, localActions, localReusableWorkflows)
-	proc.wait()
+	proc.Wait()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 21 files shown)


rhysd:
What is the reason to make these APIs public? Especially, `RuleBase` is a base type for existing rules so it is unnecessarily public if you just use the rules. Do you make your own rules?

hugo-syn:
Yes I would like to make my own rules

rhysd (changes requested):

hugo-syn:
I was actually reimplementing my own Linter to run my custom rules, because I don't want all the rules, so I just need to access the _RuleBase_ and the methods _error_ and _errorf_ to create new rules

rhysd (approved):

rhysd:
OK. But please note that you can filter out errors from specific rules by `-ignore`.

rhysd:
I added `OnRulesCreated` hook API at  https://github.com/rhysd/actionlint/commit/655d745419ce1cf2cdf94d1ddf613c8cf14dfdde.

You can add/remove rules without reimplementing `check` method.

```go
o := &LinterOptions{
    OnRulesCreated: func(rules []Rule) []Rule {
        // Remove 'runner-label' rule
        for i, r := range rules {
            if r.Name() == "runner-label" {
                rules = append(rules[:i], rules[i+1:]...)
                break
            }
        }
        // Add your own rule
        rules = append(rules, NewYourOwnRule())
        return rules
    },
}
```

So you no longer need to reimplement `Linter.check` only for adding/removing rules. With this API, can the `ConcurrentProcess` API be back to private since you no longer need to call `NewRuleShellcheck` by yourself?

I'd like to keep the API private since it is an implementation detail and can be updated on patch releases.

hugo-syn:
Sounds nice, I think that I don't need to re-implement my linter now, thanks :)

rhysd:
Great. Thank you for confirming that. I'll revert `ConcurrentProcess` to `concurrentProcess`

PR #329, opened 2023-08-02
Change Debug to public and add new script injection input

hugo-syn:
I've changed the _Debug_ method of _RuleBase_ since we can now add new rules.

And I've also added a new entry in _BuiltinUntrustedInputs_ because I already saw something like this where It's possible to inject code in the runner from an opened issue:

```yaml
name: Test

on:
  issues:
    types: [opened]

jobs:
  dummyjob:
    runs-on: ubuntu-latest
    steps:
      - name: parse issue Body
        id: prepareBody
        shell: pwsh
        run: |
          # this script parse issue and set env variables
          .\scripts\issue-body-parser.ps1 $env:BODY
        env:
          BODY: ${{ github.event.issue.body }}
      
      - name: Injection step
        shell: pwsh
        run: |
          
          # InjectionEntry comes from issue-body-parser.ps1 
          # which parse the body and do something like this
          # Write-Output "InjectionEntry=$dataparsedFromIssue" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append

          .\scripts\my-script.ps1 -param ${{ env.InjectionEntry}}
        
```

```
./main env-inject.yml       
env-inject.yml :22:265: "env.*" is potentially untrusted. avoid using it directly in inline scripts. instead, pass it through an environment variable. see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions for more details [expression]
```

### DIFF
--- a/expr_insecure.go
+++ b/expr_insecure.go
@@ -24,6 +24,9 @@ func (m *UntrustedInputMap) findObjectProp(name string) (*UntrustedInputMap, boo
 		if c, ok := m.Children[name]; ok {
 			return c, true
 		}
+		if c, ok := m.Children["*"]; ok {
+			return c, true
+		}
 	}
 	return nil, false
 }
@@ -132,6 +135,9 @@ var BuiltinUntrustedInputs = UntrustedInputSearchRoots{
 			),
 		),
 		NewUntrustedInputMap("head_ref"),
+	),
+	"env": NewUntrustedInputMap("env",
+		NewUntrustedInputMap("*"),
 	),
 }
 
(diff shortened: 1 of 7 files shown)


rhysd:
Doesn't the unit test fail on your machine? Please ensure `go test ./...` passes before making a PR on local.

rhysd:
Can you separate this PR into two PRs?

1. making `Debug` public
2. add untrusted value filter pattern

I think 1. can be easily merged but 2. would take time to fix unit tests.

PR #357, opened 2023-09-28
feat: allow the use of if condition to trigger github events

toVersus:
We could add if condition to trigger workflows with all the supported GitHub events. Please refer to the example repository below.
https://github.com/toVersus/github-actions-sandbox

However, when running actionlint on the example repository, it will return the following error:

```console
❯ actionlint
.github/workflows/action.yaml:3:5: unexpected key "if" for "branch_protection_rule" section. expected one of "branches", "branches-ignore", "paths", "paths-ignore", "tags", "tags-ignore", "types", "workflows" [syntax-check]
  |
3 |     if: true
  |     ^~~
(...)
```

This PR enables the linter to permit the use of if condition for triggering GitHub events.


rhysd:
I couldn't find documentation for this syntax on https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions. Would you show some pointer? I'd like to confirm this is officially defined by GitHub.

toVersus:
> I couldn't find documentation for this syntax in any official documents:

Unfortunately, I couldn't find the documentation either, so I demonstrated the behavior using the sample repository.
The specific use case is related to [Digger](https://digger.dev/) with [the following setup](https://docs.digger.dev/getting-started/github-actions-and-gcp#5.-create-a-workflow-file).

```yaml
name: Terraform Digger

on:
  pull_request:
    branches: [ "main" ]
    types: [ closed, opened, synchronize, reopened ]
  issue_comment:
    types: [created]
    if: contains(github.event.comment.body, 'digger')
  workflow_dispatch:

jobs:
  (...)
```

I understand that this request  may not be considered acceptable since it's not documented behavior. In that case, we will proceed by continuing to ignore the specifi linter error using the following command.

```sh
actionlint -color -ignore 'unexpected key "if" for "issue_comment" section'
```

rhysd:
Thank you for making the patch. I couldn't find any documentation which explains that `on.<event_type>` can have `if` guard. Could you show some pointer to any official document?

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on

rhysd:
https://github.com/orgs/community/discussions/108504

Issue #46, opened 2021-09-02
Feature Request: Support linting action metadata files


mike-stewart:
It would be great to have support for linting action files as well as workflow files. 

Especially with the recent release of composite actions, it seems that building private actions will be a common way to DRY up workflow files. Currently there is a gap in linting, because the workflow files are covered but anything moved to an action is not.

rhysd:
References:

- Changelog: https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/
- Document: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

steve-viso:
Hey, I was wondering if there is any plans to implement this?

krisrecur:
This would be very useful...

andyfeller:
@rhysd : truly appreciate your good work here! :sparkles: 🙇 

**if someone was interested in contributing support for this feature, what would be some considerations for adding this that someone unfamiliar with the project might review?**

1. Would `actionlint` be invoked with an arbitrary path containing composite action(s) and workflows, not simply `.github/workflows/`?
2. How difficult is it to account for a similar but different schema for composite actions versus a workflow file?

SimonHeimberg:
Some ideas:
- distinguish between workflow and action:
    - workflow has top level key `jobs`
    - action has top level key `runs`
- find action files:
    - maybe those referenced in the workflow files
    - maybe `**/action.yaml` (to find public actions)
    - maybe `.github/action*/*.yaml`
- what to check:
   - runs.using is `composite` or a node version (nodeXX) or `docker`
   - runs.steps of composite actions like steps of a job ([p.parseSteps()](https://github.com/rhysd/actionlint/blob/39d46a3b37112a04da8594c7119cf9280b779cae/parse.go#L1066)) (but [shell is required if run is set](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsshell))
   - ??? for docker actions
   - ??? for JavaScript actions
   - inputs
   - outputs
   - description exists (is mandatory)

karlbrown-va:
one thing that might help is this: https://check-jsonschema.readthedocs.io/en/stable/usage.html
```
  - repo: https://github.com/python-jsonschema/check-jsonschema
    rev: 315c23f8562a5bafa15ab448ff542abe02d9e439  # frozen: 0.28.2
    hooks:
      - id: check-github-actions
```

[...]

PR #366, opened 2023-10-17
feat: Lint actions

msw-kialo:
This PR tries to address #46 to lint action metadata files (`action.yml` / `action.yaml`) analog to workflow files. In addition to scheme validation this allows finding more sophisticated errors by running pyflake, shellcheck, validation expressions and other cross field references (e.g. outputs). The main advantage is obviously for people writing composite actions.

### Included changes

This PR tries to cover all essential aspects of this feature and provides example implementation for most of them:
* Add new `parse*` methods to parse action metadata.
* Add a new input-format parameter (both to the CLI and linteropts) to decide whether the input should be treated as workflow file or action metadata file.
* Adapt rules if needed to handle action metadata files, too.
* Add new branding rule to validate branding values.
* Adapt the playground to allow linting of actions, too.
* Add tests to correctly valid common actions and caught various issues.

### Implementation decisions

We needed to take a few implementation decisions where it is not clear whether to picked the best option (once they are decided we will squash our changes and mark the PR as ready):

* **API stability**: the public interface was kept unchanged unchanged (if needed function alias like `Parse` and `Visit` were added). At the moment, the visitors `pass` interface is extended by `VisitActionPre` and `VisitActionPost`, however it reuses `visitStep`. This could "crash" (nil value accessed) custom rules as `VisitWorkflowPre`, `VisitJobPre` might not be called before `visitStep`: the `RuleID`, `RuleExpression` had this issue.
* **parseAction vs ActionMetadata**: The local action metadata cache (`action_metadata.go`) was left unchanged (it appears to be more light-weight and overall simpler to leave as small duplication).
* **LintDir behavior**: to keep the change simple, `LintDir` will only lint arbitrary `.yml`, `.yaml` files if the directory is `.github/workflows`. This should be an issue in practice (as workflows needs to be in such directory to be interpreted by GitHub), but it required the `testdata/projects` tests to be adapted (specify `-input-format=workflow` would also work).
* **Default behavior**: actions are linted by default (at least if `actionlint` is invoked without extra arguments)
* **Action location**: At the moment, any `action.yml` / `action.yaml` file within the repository is considered an action metadata file (top level, or in any subdirectory at any level). While it ensures by default any such file is discovered, it might slow if it is a big repository and could cause false positives (the filename isn't that specific). Discovery could be reduced; but, new config option might be needed to handle more complex use-cases. `-input-format=workflow` was added to the dog-food CI calls to ensure the `testdata` actions aren't discovered.
* **CLI shortcut**: this PR added `actionlint $dir` to lint only files within the specific directory, e.g. `actionlint .github/actions` (it was originally added for testing purposes but it might be helpful for others, too).
* **Expression values**: Sadly, we haven't found anything in the GitHub documentation which expressions are supported action metadata files (probably the step files are the same, however, `inputs.X.default`, `runs.envs.X`, `runs.pre-if`, `runs.post-if` are new. E.g. I suspect special functions like `always()`, `success()` are supported for `runs.pre-if` but I can't proof it.
* **Branding values**: the allowed branding values (`color` and `icon`) are currently hardcoded. I suspect, a script to extracted them from GitHub's docs should be added?


msw-kialo:
Marking PR as ready to indicate further work depends on PR review.

Ezri-Mudde:
What's the status on this? Will this eventually be merged or should I look for another tool to lint action files? I've tried @msw-kialo changes on my action files and it seems to work perfectly. I would like to use this in my actions workflow without having to clone and build their fork.

msw-kialo:
I am ready to make needed adjustments, rebases, restructure the PR/change etc.
However, I am waiting for a first review from @rhysd before I invest more time in a potentially wrong direction.

ChrisCarini (commented):
Overall, LGTM! 🎉 👏 

Left several suggestions.

I also tested this branch locally against ~10 workflows and ~50 actions after I merged in the latest changes (`kialo:lint-actions` is about 250 commits behind `main` right now; a few trivial merge-conflicts to resolve).

ChrisCarini:
> - **Branding values:** the allowed branding values (color and icon) are currently hardcoded. I suspect, a script to extracted them from GitHub's docs should be added?

I'm not sure if GitHub themselves provides anything, but there is a JSON payload that has the respective values here: https://github.com/rhysd/actionlint/pull/366

Honestly, I don't think these change often enough that it warrants the added complexity in this initial PR. My 2 cents is this can be tracked as a future feature improvement, and implemented if/when it becomes an issue for folks down the road.

ChrisCarini:
> @ChrisCarini Thanks for testing it and providing feedback.

Hi @msw-kialo - no problem! Thank you for implementing this functionality and posting the PR! 🙏 

> I am still open to make needed changes to this PR. However, since https://github.com/rhysd/actionlint/releases/tag/v1.7.0: ...
> 
> It appears that @rhysd has other architecture plans. So until I hear where this is going I won't invest more time into this. Actually, I consider just closing this PR as it is apparently not appreciated.

I understand that's a bit frustrating :( FWIW, I would actually really love it if your fork for this PR was released separately until this repo either (a) decides to provide further comment in pursuit of getting the feature merged in, or (b) decides to implement this support some other way.

Ultimately, I really just want this functionality (and personally don't mind who/where it is implemented and released) so I can start using faster.

Sam13:
Any updates on that?
Just lining workflows is not enough ;-)

PR #375, opened 2023-11-01
feat: add inline-ignore

### LINKED ISSUE #237: Inline ignores in files
I'd love to see inline/annotation support for ignore rules. Many other linters (one example: shellcheck) does the same.

The ignore rule would then only apply to the following line instead of all files in the run. Similarly, applying it to the top of the file would only apply it to said file.

Example (from actionlint annotation examples):

```yaml
      - name: Run Actionlint
        # actionlint ignore=SC2016
        run: |
          actionlint -format '{{range $err := .}}::error file={{$err.Filepath}},line={{$err.Line}},col={{$err.Column}}::{{$err.Message}}{{end}}' .github/workflows/*.yml
```

kounoike:
add inline-ignore feature.
this resolves #237

### DIFF
--- a/linter.go
+++ b/linter.go
@@ -500,7 +500,7 @@ func (l *Linter) check(
 		l.debug("No config was found")
 	}
 
-	w, all := Parse(content)
+	w, i, all := Parse(content)
 
 	if l.logLevel >= LogLevelVerbose {
 		elapsed := time.Since(start)
@@ -592,6 +592,24 @@ func (l *Linter) check(
 			for _, pat := range l.ignorePats {
 				if pat.MatchString(err.Message) {
 					continue Loop
+				}
+			}
+			filtered = append(filtered, err)
+		}
+		all = filtered
+	}
+
+	// ignore inline ignore patterns
+	if len(i) > 0 {
+		filtered := make([]*Error, 0, len(all))
+	Loop2:
+		for _, err := range all {
+			patterns, ok := i[err.Line]
+			if ok {
+				for _, p := range patterns {
+					if p.MatchString(err.Message) {
+						continue Loop2
+					}
 				}
 			}
 			filtered = append(filtered, err)
--- a/parse.go
+++ b/parse.go
@@ -1402,14 +1402,41 @@ func handleYAMLError(err error) []*Error {
 	return []*Error{yamlErr(err.Error())}
 }
 
+var ignoreRe = regexp.MustCompile("(?:^|\n)\\s*#\\s*actionlint\\s+ignore=([^\n]+)\\s*$")
+
+func visitNodeForIgnores(n *yaml.Node, i *map[int][]regexp.Regexp) {
+	if n.HeadComment != "" {
+		line := posAt(n).Line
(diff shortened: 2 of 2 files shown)


vl-kp:
someone please merge this PR

rhysd:
This PR is not acceptable because

- no test
- no document
- it cannot ignore errors in the middle of a multi-line string like below
  ```yaml
  - run: |
      ${{
        hoge
      }}
  ```

The feature requested at #237 is more complicated than it looks actually.

kounoike:
@rhysd Could you merge this PR?

- I wrote test
- I wrote document
- it **can** ignore error in multi-line string

```yaml
        # actionlint ignore=potentially untrusted
        run: |
          echo "hello"
          echo '${{ github.event.head_commit.author.name }}'
          echo "hello"
```

I hope to have a constructive discussion.

rhysd:
Thank you for addressing some of my comments.

> it can ignore error in multi-line string

That's possible because the error position is poor. For example,

```
L10 - run: |
L11    ...
L12    ...
L13    ${{ hoge }}
```

When an error is found at L13, actionlint reports an error happened at L10. So implementation in this branch somehow currently works fine and the ignore command can remove the error.

However, when we improve the error position in the future and actionlint can report an error happened at L13, then this implementation will break. The ignore comment above L10 cannot remove an error at L13. And there is no possible fix in that case.

So this implementation happens to work correctly for now, but it's depending on poor error location. Once the poor error location issue is solved, this implementation no longer works.

This is what I concern.

ChrisCarini:
@rhysd I was wondering about the 'poor error location' when I've been using this tool - is there a separate issue tracking that bug/improvement? I'd certainly love to see this improve.

My 2 cents on this PR, if this works w/ the current way the error location is reported, I think this is fine, and when the poor error location problem gets addressed, the implementation in this change can be updated to accommodate it as well - what do you think?

thiagowfx:
Is there another way to disable checks in the meantime?

For example, I got a false positive due to github composite actions:

```
.github/workflows/foo.yml:31:31: property "test_server_ip" is not defined in object type {} [expression]
```

The input exists in the composite action but `actionlint` cannot recognize it (https://github.com/rhysd/actionlint/issues/401). I cannot seem to find a way to disable the check in the meantime.

gringolito:
It’s been nearly a year since this PR was opened and ultimately rejected due to an issue that may or may not have been addressed since then. Has the problem with the 'poor error location' been resolved in the meantime?

If not, I believe it would make sense to reconsider this PR, as it offers a practical solution that could benefit many users who are still waiting on this feature.

ChrisCarini:
Hey @rhysd - what are your current thoughts on this PR in late-2025? I recall you expressed concerns about the error location on yaml multi-line blocks - is there a separate issue tracking that?

I'd like to encourage this PR move forward based on the current functionality of actionlint, and not be held up on a future improvement. I'd love to see this PR merged in to be able to be used by the men's who are eager to use it!

WDYT?

Best, 
Chris

netflash:
Hi @rhysd!
If @kounoike or myself fix the conflicts, would you be willing to merge this one?

Issue #377, opened 2023-11-08
Support glob/regex patterns for self hosted runner labels configuration


kishaningithub:
## Context
To add self hosted runners to the list i am doing something like

```yaml
self-hosted-runner:
    labels:
      - private-runner-label1
      - private-runner-label2
```

but the problem the above way of configuration is not ideal if the label contains a dynamic elements, For example 

```yaml
self-hosted-runner:
    labels:
      - INSTANCE_TYPE=m6a.large # Yes this is a label :-)  
```

In the above the value of the INSTANCE_TYPE can be anything (t3.large, m2.large etc etc) and they are all valid github runner labels

## Proposal

It would be great if this configuration can be made into a pattern like

```yaml
self-hosted-runner:
    labels:
      - INSTANCE_TYPE=*  # The * matches everything  
```

This way i don't have to specify each and every label

rhysd:
Closing as #378. Now glob pattern is supported.

Issue #359, opened 2023-09-28
[feature] self-hosted-runner pattern/regex


jenseng:
I'm using self-hosted runners, but the labels are not statically knowable, as they can include dynamic parameters, e.g.

```yaml
runs-on: [self-hosted, "some-base-prefix:size=large&cpu=8"]
```

While it looks like I could either turn off the `runner-label` rule, and/or define my own rule with `OnRulesCreated` and use actionlint programmatically, it would be awesome if actionlint itself had some more flexibility around how to lint these labels. Not necessarily fully exhaustive, just being able to do a regex or glob style pattern would be great.

dekimsey:
This would be great, I'd love to have some glob support here as we run into the same issue ourselves!

kishaningithub:
For some reason this issue did not come up in the search and i created a similar ticket. Linking it here

https://github.com/rhysd/actionlint/issues/377

kishaningithub:
@jenseng Let know if my proposal #377 would also work for you

kishaningithub:
@jenseng / @dekimsey Have created a PR address this https://github.com/rhysd/actionlint/pull/378 feel free to put in your feedback in the PR as well

jenseng:
This is great, thanks @kishaningithub!

MPV:
For ideas/context, here's someone else doing the equivalent of this (in a related kind of solution):
- https://github.com/mpalmer/action-validator/issues/66

rhysd:
Closing as #378. Now glob pattern is supported.

PR #378, opened 2023-11-08
Add glob support for self hosted runner known labels configuration

kishaningithub:
Addresses #359 and #377

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@
 /man/actionlint.1.html
 /playground-dist
 /actionlint-workflow-ast
+# IntelliJ IDE Folder
+.idea/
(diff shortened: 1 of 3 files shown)


dekimsey:
Thank you for this @kishaningithub, this will address our ability to lint with custom runners! 👍

.gitignore
```diff
@@ -12,3 +12,5 @@
 /man/actionlint.1.html
 /playground-dist
 /actionlint-workflow-ast
+# IntelliJ IDE Folder
+.idea/
```
jenseng, line 16: I'll let the maintainers chime in, but typically I'd put this sort of thing in my own global gitignore; people develop in all kinds of environments, so adding that kind of stuff to a project-level gitignore can make it unwieldy

rhysd (approved):
Nice addition. Thank you.

PR #381, opened 2023-12-03
Add the Arch Linux official repository in the installation documentation

sorairolake:
In addition to the AUR, [`actionlint`](https://archlinux.org/packages/extra/x86_64/actionlint/) is also available in the Arch Linux official repository.


rhysd (approved):
Thanks!

PR #382, opened 2023-12-07
pre-commit: include fully qualified container name

gotmax23:
Alternative container engines such as podman don't default to Docker Hub and require specifying the fully qualified container url.


rhysd (approved):

rhysd:
Thanks!

PR #396, opened 2024-02-09
Remove --norc flag from shellcheck command args

### LINKED ISSUE #395: Feature request: Allow usage of .shellcheckrc
Thanks for this great tool. 

There seems to be one issue that kind of blocks adapting this tool fully into existing codebase. It seems actionlint does not use `.shellcheckrc` file at all because of this `--norc` in here https://github.com/rhysd/actionlint/blob/3a2f2c755b6442ec7999b28dfd107e1bb9853389/rule_shellcheck.go#L172 
Would you consider removing this flag? Would this make `shellcheck` load file we have in project? Thanks

piotrekkr:
This should allow for shellcheck to load custom configuration if exist in project root.

fixes #395


piotrekkr:
Closing because of https://github.com/devops-actions/actionlint/issues/40

PR #400, opened 2024-02-19
feat: Add `workflow_run.workflows` references check

### LINKED ISSUE #83: Validate workflow_run's workflow names actually exist
One of the most common errors I encounter with our workflow files is when we reference other jobs in `needs` and have a typo there or update job names while forgetting to update the dependent jobs.
Luckily actionlint is able to find this kind of error with the job-needs check.

Is there any possibility to have that kind of check for `on.workflow_run.workflows` to verify whether the specified workflow name actually exists in the project?

gmazzo:
Fixes #83 by introducing a new `workflow_run` rule to check the existence of referenced workflows at that trigger.

This rule is a bit different than others, it will be a singleton instance for the check run (instead of one per file), to allow collecting workflow names from the whole suite.

Because of that, it can only apply when calling `LintFiles` (`nil` for other usages of the linter).
It validates that any `name` reference on `workflow_run.workflow` exists in another workflow of the same set.

```yaml
on:
  workflow_run:
    types: [completed]
    workflows:
      - build
      - test # ERROR: This workflow does not exist, its name is `Tests`
```
image

This is my first experience working with Go, feel free to suggest any change regarding code style or best practices.

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -897,6 +897,8 @@ type Job struct {
 // Workflow is root of workflow syntax tree, which represents one workflow configuration file.
 // https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
 type Workflow struct {
+	// Path is the path of the workflow. This field is internal and not defined by GitHub workflow's syntax
+	Path string
 	// Name is name of the workflow. This field can be nil when user didn't specify the name explicitly.
 	// https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#name
 	Name *String
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -36,6 +36,7 @@ List of checks:
 - [Environment variable names](#check-env-var-names)
 - [Permissions](#permissions)
 - [Reusable workflows](#check-reusable-workflows)
+- [Workflow references at `workflow_run`](#check-workflow-run-refs)
 - [ID naming convention](#id-naming-convention)
 - [Contexts and special functions availability](#ctx-spfunc-availability)
 - [Deprecated workflow commands](#check-deprecated-workflow-commands)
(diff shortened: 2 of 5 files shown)


pass.go
```diff
@@ -15,7 +15,7 @@ type Pass interface {
 	// VisitJobPost is callback when visiting Job node after visiting its children. It returns internal error when it cannot continue the process
 	VisitJobPost(node *Job) error
 	// VisitWorkflowPre is callback when visiting Workflow node before visiting its children. It returns internal error when it cannot continue the process
-	VisitWorkflowPre(node *Workflow) error
+	VisitWorkflowPre(path string, node *Workflow) error
```
gmazzo, line 18: I need the `path` here to infer the name from the filename, because it's not always defined explicitly on the workflow's YAML.

I also explored adding a new `VisitFile` method to avoid touching many failed to match the new `VisitWorkflowPre` signature, but because the stateful nature of this rule, and the concurrent approach for processing files, the current name stored as a field of the rule itself gets overridden, leading to a race condition.

So I end up adding this extra parameter here to compute the final name.

I could also add the final name to the `Workflow` structure, but it doesn't feel right, as that stands for a pure YAML model.

gmazzo:
Hi @rhysd, I see there is activity in this repo. Any chance to get this one reviewed? We are currently using a forked internal version of your tool.

Great work by the way!

gmazzo:
Hi @rhysd, I've rebased the code, simplified the solution and now all tests seem to be OK. It would be nice to hear back from you, this has been ignored for a while now :(

gmazzo:
It has been 9 months and this PR addressing #83 it has been just ignored. We'll keep using out forked version then 🤷‍♂️

PR #413, opened 2024-04-19
Update `checkout` action version in documentation

naglis:
`actions/checkout@v3` produces a warning about deprecated Node.js version ([see also](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)).


rhysd (approved):
Thanks for catching this.

PR #422, opened 2024-05-10
preactjs/compressed-size-action@v2 is new

ne-sachirou:
preactjs/compressed-size-action@v2 was old to points 2.5.0. Now it's new to points 2.6.0.

https://github.com/preactjs/compressed-size-action/tags

v2 (2.6.0) uses node20 runtime.

https://github.com/preactjs/compressed-size-action/blob/v2/action.yml#L46

### DIFF
--- a/popular_actions.go
+++ b/popular_actions.go
@@ -3770,7 +3770,6 @@ var OutdatedPopularActionSpecs = map[string]struct{}{
 	"peter-evans/create-pull-request@v2":               {},
 	"peter-evans/create-pull-request@v3":               {},
 	"preactjs/compressed-size-action@v1":               {},
-	"preactjs/compressed-size-action@v2":               {},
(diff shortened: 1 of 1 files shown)


rhysd:
I'm not understanding what is your issue. Can you elaborate on it? `v2` tag is dynamically updated.

rhysd:
OK, I think I could understand the issue.

- v2.5.0 uses `node12`
- v2.6.0 uses `node16`
- v2 was recently updated from v2.5.0 to v2.6.0 (probably within a few days? since popular_actions.go was updated 3 days ago)

So v2 is still marked as outdated in popular_actions.go

rhysd (changes requested):
This change is not acceptable since popular_actions.go is generated from script and should not be modified directly as the header of the file says.

I'll update it on my side.

PR #451, opened 2024-09-27
Remove `macos-11` and `macos-11.0` runner labels support

muzimuzhi:
`macos-11` runner label was dropped in Jun 2024.
https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/#macos-11-deprecation-and-removal

I think `macos-11.0` runner label was dropped too (see https://github.com/actions/runner-images/issues/10099#issuecomment-2317454074), though [another GitHub blog post](https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/) said it is not deprecated until Dec 3, 2024.


muzimuzhi:
actions/runner-images@22143c7c ([macOS] Remove macOS-11 related code due to the deprecation, 2024-08-09) removed the code to build the macOS 11 image a month ago, so yes `macos-11.0` is indeed dropped.

rhysd (approved):
Thank you for catching this. Looks good.

PR #454, opened 2024-10-02
feature: warn when reusable workflow is used as a step

IlyaGulya:
Using reusable workflow as a step is not supported by Github Actions.
However, actionlint currently does not warn about this behavior.

Now it does 🙂

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -165,6 +165,8 @@ paths:
     ignore: [vvv]
   .github/workflows/**/x.yaml:
     ignore: [www]
+  .github/workflows/**/*.{yml,yaml}:
+    ignore: [ttt]
 `
 
 	var cfg Config
@@ -197,6 +199,8 @@ paths:
 		{".github/workflows/a/b/x.yaml", "www", true},
 		{".github/workflows/a/b/c/x.yaml", "www", true},
 		{".github/workflows/a/b.yaml", "this is not ignored", false},
+		{".github/workflows/a.yml", "xxx", false},
+		{".github/workflows/a.yml", "ttt", true},
 	}
 
 	for _, tc := range tests {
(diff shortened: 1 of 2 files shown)


rule_action.go
```diff
@@ -323,6 +323,11 @@ func (rule *RuleAction) VisitStep(n *Step) error {
 
 	spec := e.Uses.Value
 
+	if rule.isReusableWorkflow(spec) {
```
aparfenovv, line 326: в гошке тоже экстеншены есть, приколько

rule_action_test.go
```diff
@@ -0,0 +1,71 @@
+package actionlint
+
+import (
+	"testing"
+)
+
+func TestRuleAction_ReusableWorkflow(t *testing.T) {
```
aparfenovv, line 7: nit: аргумент бы назвать поточнее
`testRuleAction`?

aparfenovv (approved):

vsbauer-ind (approved):

rhysd:
I'm not sure what this feature tries to solve. Is it a common mistake to pass a workflow YAML file to `uses`? How did you confirm that it's common?

IlyaGulya:
For context, there's examples of people asking about this:
https://stackoverflow.com/questions/75733616/github-actions-how-to-call-a-reusable-workflow-as-a-step
https://www.reddit.com/r/github/comments/155yvdf/reusable_workflow_as_steps_for_caller_workflow/

It's not common, though 🙂

rhysd:
Okay, I'm not against this feature but I'd like to make false positives as minimal as possible.

One possible false positive is local actions under `.github/workflows/` directory. When local actions are put in the directory, for example `.github/workflows/actions/my-action`, they are used like

```yaml
- uses: ./.github/workflows/actions/my-action
  with:
    arg: ...
```

This check will report the valid action usage as invalid reusable workflow usage.

IlyaGulya:
Sure, I will fix that

PR #457, opened 2024-10-12
Skip PR creation in the generate workflow in forks

muzimuzhi:
As it's unnecessary (as the needed PR will be opened by workflow runs in this, the original "upstream" repository) and only causes noise in forks. See for example [this run](https://github.com/muzimuzhi/actionlint/actions/runs/11305433070/job/31444920812#annotation:7:100) in my fork.


rhysd (approved):
Good catch. Thanks!

PR #458, opened 2024-10-16
Fix CI, in another way

muzimuzhi:
The GitHub Docs source has taken the `<code>...&lt;...&gt;...</code>` to `` `...<...>...` `` changes 4 months ago, see https://github.com/github/docs/commit/a15d2618200b0567461d47bb3994d6378ca4c7b8.


rhysd:
Thank you for taking time to look into this.

However this is not a fix but a workaround not to make CI fail. The root cause is that goldmark has changed the behavior of `Node.Text`.

https://github.com/yuin/goldmark/issues/471

This PR modifies the test data not to hit this issue. So after merging this PR, the issue is still there. You'll find that `Node.Text` calls cause bunch of warnings from staticcheck due to deprecation if you update goldmark to v1.7.8.

Since I already know the reason, I'll add a fix for this.

rhysd:
I used the idea of 0c5f33d at 763f8866bb35f1775e051827c5dbc77b949e2963 so I credit you at the commit.

muzimuzhi:
Thank you. 😄

PR #459, opened 2024-10-17
Fix false-positive script injection warnings when using builtin functions with stable outputs

IlyaGulya:
# Problem
Currently there are cases when `actionlint` falsely assumes expression to be unsafe when used inside scripts and warns about possible script injection.

For instance, following step description:
```yaml
- name: Determine if release branch
  run: |
    echo "is_release_branch=${{ contains(github.head_ref, 'release') }}" >> $GITHUB_OUTPUT
```
would lead to warning:
```
.github/workflows/test.yaml:17:51: "github.head_ref" is potentially untrusted. avoid using it directly in inline scripts. instead, pass it through an environment variable. see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions for more details [expression]
```

# Why do I assume this to be false positive
The only place which I suppose relevant to the issue is this: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections
Thus, if the point of this lint rule is to ensure that script injection is not possible, example above is definitely false positive.
It's false positive because there's no way for attacker to get some arbitrary output from the `contains` function.

# What is done in this PR?

1. I've added a `parent` field to every expression AST node
2. I'm using this `parent` field to traverse the AST and search for `FunctionCallNode` parents
3. If parent is found and it's `Callee` considered safe for script injection attacks - we skip all further insecure input checks in this AST leaf.

# Which functions considered safe?
Basically, I've chosen only functions which are returning booleans. Maybe there's more, but I decided to start with these:

- `contains`
- `startsWith`
- `endsWith`

# P.S.
If there's a better way to implement this - I would love to apply any fixes required for this PR to get into `main`.
I've added the `parent` to not pollute the calls to `sema.check` with additional argument just for this case.
I hope additional memory usage would not be that high, since it's just an additional pointer per AST node.

### DIFF
--- a/expr_insecure_test.go
+++ b/expr_insecure_test.go
@@ -193,14 +193,14 @@ func TestExprInsecureDetectUntrustedValue(t *testing.T) {
 			},
 		},
 		testCase{
-			"contains(github.event.pages.*.page_name, github.event.issue.title)",
+			"fromJson(github.event.pages.*.page_name, github.event.issue.title)",
 			[]string{
 				"github.event.pages.*.page_name",
 				"github.event.issue.title",
 			},
 		},
 		testCase{
-			"contains(github.event.*.body, github.event.*.*)",
+			"fromJson(github.event.*.body, github.event.*.*)",
 			[]string{
 				"github.event.",
 				"github.event.",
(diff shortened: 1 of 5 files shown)


rhysd:
Thank you for this patch. It sounds a good addition. I'll take a look at more details later.

expr_insecure_test.go
```diff
@@ -192,20 +192,6 @@ func TestExprInsecureDetectUntrustedValue(t *testing.T) {
...
-				"github.event.issue.title",
-			},
-		},
-		testCase{
-			"contains(github.event.*.body, github.event.*.*)",
-			[]string{
-				"github.event.",
-				"github.event.",
-			},
-		},
```
rhysd, line 208: Instead of removing these tests, would you use some other unsafe functions instead of `contains`? These cases test insecure properties usage inside function calls.

rhysd (changes requested):

expr_insecure_test.go
IlyaGulya: Sure, replaced `contains` with `fromJson`

rhysd (approved):
I may do some followup after merging this PR, but changes basically look good to me. Thanks!

PR #463, opened 2024-10-21
add gcr.io and pkg.dev to valid docker regstry domain

takaidohigasi:
## WHAT

SSIA

## Note

gcr.io and pkg.dev is docker registry domain for GCP

* gcr.io: 
https://cloud.google.com/container-registry/docs/overview
* pkg.dev
https://cloud.google.com/artifact-registry/docs


rhysd (approved):
Thanks. It might be not possible to check Docker registries reasonably, but let us go with these checks for now.

PR #465, opened 2024-10-21
remove `macos-12` runner label

shogo82148:
The macOS 12 runner image has been deprecated on 10/7/2024, and will be fully unsupported by 12/3/2024

- https://github.com/actions/runner-images/issues/10721
- https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/

### DIFF
--- a/rule_runner_label.go
+++ b/rule_runner_label.go
@@ -12,9 +12,6 @@ const (
 	compatUbuntu2004 runnerOSCompat = 1 << iota
 	compatUbuntu2204
 	compatUbuntu2404
-	compatMacOS120
-	compatMacOS120L
-	compatMacOS120XL
 	compatMacOS130
 	compatMacOS130L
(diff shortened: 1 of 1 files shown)


rhysd:
Thanks. I believe we should merge this after 2024 12/3 and after this issue has been closed.

https://github.com/actions/runner-images/issues/10721

rhysd (changes requested):
This review comment is to block the merge until the runner is completely dropped.

AlexSkrypnyk:
Also, `macos-15` is available, but reported as false positive by actionlint.

rhysd:
`macos-15` was supported at 1.7.3. If actionlint still complains about it, it should be a bug.

https://github.com/rhysd/actionlint/releases/tag/v1.7.3

PR #466, opened 2024-10-25
docs: add SC2043 to disabled shellcheck rules

risu729:
Following up this commit.
https://github.com/rhysd/actionlint/commit/c71a576bdf418ffa9cc06b8f7d0cb7a500fa4846


rhysd (approved):
Thanks! I missed it.

Issue #469, opened 2024-11-06
Check whether a specific action has been used


chrisreddington:
I'm exploring options to determine a 'baseline' of the GitHub Actions used in workflows across a set of repositories, for example - for example, the lint step should fail if a workflow does not contain `chrisreddington/my-action` (regardless which version of the action is used).

There could be two different checks:

- A check which determines whether _**any**_ version of the action is used
- A check which determines whether _**a specific**_ version of the action is used

Then, I'd be able to use `actionlint` within a reusable workflow to determine whether a set of repositories conform to some baseline expectations.

chrisreddington:
I've taken an attempt at raising a PR for this, but it's my first time deeply looking at this codebase so I suspect I'll need some changes in the PR :)

There's one bit of the implementation that I think could also use some 👀 and a second opinion in terms of how it handles reviewing multiple jobs in a workflow. Whether it should just check for existence within the workflow, or within each individual job as well... I'm not sure I have a strong opinion on which way to go right now.

PR #474, opened 2024-11-11
Add required actions rule (and supporting tests) to enforce usage of specific GitHub Actions in workflows

chrisreddington:
This PR implements issue #469 that I raised a few days ago. It introduces a new linting rule to ensure specific GitHub Actions are present in workflows, along with corresponding tests. Changes include updating the configuration structure, adding the new rule to the linter, implementing the rule, and creating tests for it.

### Configuration Changes:
* Added `RequiredActions` field to the `Config` struct in `config.go` to specify mandatory GitHub Actions in workflows.

### Linter Rule Implementation:
* Updated `linter.go` to include the new rule `RuleRequiredActions` if `RequiredActions` are specified in the configuration.
* Created `rule_required_actions.go` to define the `RequiredActionRule` struct and implement the `RuleRequiredActions` class, which checks for the presence and correct version of required GitHub Actions in workflows.
  * If the Action is present, then it checks for any version of the Action.
  * If the Action is present as well as a specific version, then it checks for a specific version of the Action. 

### Testing:
* Added `rule_required_actions_test.go` to test the functionality of `RuleRequiredActions` and the `parseActionRef` function. This includes various test cases to ensure the rule works correctly under different scenarios.


mifriis:
@chrisreddington this is a very interesting addition.

I am investigating if we can introduce some light governance where certain actions are required to be part of our enterprises workflows. It sounds like this approach could help us lint the workflow to be compliant with such requirements?

The PR has been stalled for a while. Is anyone waiting for something?

PR #479, opened 2024-11-18
Make Actionlint Directly Invokable as an Action Itself

### LINKED ISSUE #262: Make Actionlint Directly Invokable as an Action Itself
Currently, there are two ways of running `actionlint` in a GitHub Action workflow:
1. Running the download script to install `actionlint` and execute it from a `run:` step:
   ```yaml
   - name: Check workflow files
     shell: bash
     run: |
       bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
       ./actionlint -color
   ```
2. Running `actionlint` from a docker container
   ```yaml
   - name: Check workflow files
     uses: docker://rhysd/actionlint:latest
     with:
       args: -color
   ```

Both of these two options have some drawbacks. Option 1 requires either always running the download script from the main branch (which may be unstable), or specifying a stable tag (which needs to be manually updated when new versions are released as Dependabot won't suggest updates). Option 2 has the benefit of being tied to a stable version, e.g. `docker://rhysd/actionlint:v<TAG>`, however, it can only be configured by passing in the `args:` property as a monolithic string, instead of separate inputs with clear defaults and descriptions.

One solution would be to add a `action.yml` file to the root that defines `inputs:` and executes `actionlint` internally. This could be done as a composite action (as was proposed in #257) or as a dockerized wrapper (which would forgo the need to run the download script). Essentially both of these would be wrapping up one or the other of the existing methods to run `actionlint` from a workflow. The key benefits here are clearly defined inputs and enabling Dependabot updates.

The possible downsides to adding an `action.yml` file, as pointed out by @rhysd in #257, is that this adds maintenance costs and is identical to just running one of the two existing methods.

> Honestly I'm not seeing much difference between running the download script then executing the downloaded binary with `run:` and running the dedicated action with `uses:` though making a new action increases maintenance cost.

However, I believe that having a first party action would still add value. In addition to the aforementioned maintenance improvements to end users by enabling Dependabot and distinct `inputs:`, this would also add `actionlint` to the GitHub actions marketplace.

norwd:
Currently, there are two ways of running `actionlint` in a GitHub Action workflow:
1. Running the download script to install `actionlint` and execute it from a `run:` step:
   ```yaml
   - name: Check workflow files
     shell: bash
     run: |
       bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
       ./actionlint -color
   ```
2. Running `actionlint` from a docker container
   ```yaml
   - name: Check workflow files
     uses: docker://rhysd/actionlint:latest
     with:
       args: -color
   ```

Both of these two options have some drawbacks. Option 1 requires either always running the download script from the main branch (which may be unstable), or specifying a stable tag (which needs to be manually updated when new versions are released as Dependabot won't suggest updates). Option 2 has the benefit of being tied to a stable version, e.g. `docker://rhysd/actionlint:v<TAG>`, however, it can only be configured by passing in the `args:` property as a monolithic string, instead of separate inputs with clear defaults and descriptions.

This pull request solves this by adding an `action.yml` file to the root that defines `inputs:` and executes `actionlint` internally. Essentially wrapping a run `actionlint` from a workflow. The key benefits here are clearly defined inputs and enabling Dependabot updates.

This resolves #262


rhysd:
Adding an action to this repository is confusing because tags of this repository are for `actionlint` command line tool. For example, when adding a new feature to the action causes a minor version bump even if the `actionlint` command has no new features.

Currently actionlint can be used with only two steps; download it and just run it.

https://github.com/rhysd/actionlint/blob/main/docs/usage.md#use-actionlint-on-github-actions

I don't feel just converting the two steps into one is a compelling argument to the problem I explained above.

norwd:
> Adding an action to this repository is confusing because tags of this repository are for `actionlint` command line tool. For example, adding a new feature to the action causes a minor version bump even if the `actionlint` command has no new features.
> 
> Currently actionlint can be used with only two steps; download it and just run it.
> 
> https://github.com/rhysd/actionlint/blob/main/docs/usage.md#use-actionlint-on-github-actions
> 
> I don't feel the benefit of just converting the two steps into one is a compelling argument against the problem I explained above.

That's fair 🙂, I just thought since I had a fork lying around I might as well offer to merge it. Feel free to close this and the linked issue.

rhysd:
Keeping the fork or creating a separate repository for it sounds much better than merging this. Thank you for making the action :+1:. Then closing.

fregante:
@rhysd I'm confused as to why a tool that lints actions isn't itself available as an action. I'd very much rather have a first party product than install someone else's (which might end up forgotten/unmaintained)

This could be one line of code without having to download a docker image:

```yml
- uses: rhysd/actionlint
```

PR #482, opened 2024-11-21
Add `shellcheck` dependency in pre-commit hooks

### LINKED ISSUE #477: pre-commit hook does not install shellcheck, resulting in inconsistent linter behavior
Consider the [attached GitHub Actions workflow definition](https://github.com/user-attachments/files/17739777/ci.zip).

If shellcheck is not installed on the developer's computer, the actionlint pre-commit hook will not report any issues with the workflow definition.

Furthermore, if one uses a tool like [act](https://nektosact.com/) to run that workflow locally, the actionlint pre-commit hook executed within the local GitHub Actions runner will not report any issues with the workflow.  This is because [runner images commonly used with act](https://github.com/catthehacker/docker_images) do not have shellcheck installed.

However, when GitHub Actions itself runs the workflow, the linter will report these issues because [the ubuntu-latest image includes shellcheck](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md):

```
.github/workflows/ci.yml:40:9: shellcheck reported issue in this script: SC2086:info:1:106: Double quote to prevent globbing and word splitting [shellcheck]
   |
40 |         run: |
   |         ^~~~
.github/workflows/ci.yml:157:9: shellcheck reported issue in this script: SC2102:info:2:17: Ranges can only match single chars (mentioned due to duplicates) [shellcheck]
    |
157 |         run: |
    |         ^~~~
.github/workflows/ci.yml:157:9: shellcheck reported issue in this script: SC2006:style:3:11: Use $(...) notation instead of legacy backticks `...` [shellcheck]
    |
157 |         run: |
    |         ^~~~
```

This behavior is unexpected.  No matter where one runs the actionslint pre-commit hook, it should have the same result.  The hook should install shellcommit (plus any other software on which it depends) in its environment.

kadykov:
Fixes #477 

This pull request adds the `github.com/wasilibs/go-shellcheck/cmd/shellcheck@latest` package as a dependency for actionlint `pre-commit` hooks. This change resolves the issue reported in #477.

Please note that I'm not familiar with Go and may not have chosen the most optimal dependency package. If you have a better approach, please let me know and I'll be happy to update the code.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -7,7 +7,7 @@
   files: ^\.github/workflows/
   entry: actionlint
   minimum_pre_commit_version: 3.0.0
-  additional_dependencies: ["shellcheck"]
+  additional_dependencies: ["github.com/wasilibs/go-shellcheck/cmd/shellcheck@latest"]
 - id: actionlint-docker
   name: Lint GitHub Actions workflow files
   description: Runs actionlint Docker image to lint GitHub Actions workflow files
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -7,6 +7,7 @@
   files: ^\.github/workflows/
   entry: actionlint
   minimum_pre_commit_version: 3.0.0
+  additional_dependencies: ["shellcheck"]
 - id: actionlint-docker
   name: Lint GitHub Actions workflow files
   description: Runs actionlint Docker image to lint GitHub Actions workflow files


rhysd:
Since I'm not a pre-commit user, I can't say this is the right direction or not. However let me confirm one thing. Why do you use the WASI port of shellcheck instead of the [official released binaries](https://github.com/koalaman/shellcheck/releases)?

anuraaga:
> It's unclear how it will be maintained (e.g. following the new releases)

Randomly noticed this PR and just wanted to answer this point, there is an automatic update script for catching new upstream releases

https://github.com/wasilibs/go-shellcheck/blob/main/.github/workflows/update.yaml

shellcheck is the latest project but there are several other ones in wasilibs and it's been pretty easy to keep up with releases thanks to the automated updates.

That being said, I agree with the sentiment of sticking to official repos so if there's an easy way to use the official binaries with pre-commit, it's probably better, just wanted to clarify the above point.

rhysd:
@anuraaga Thank you for the explanation. Then I'm okay to rely on go-shellcheck.

However I'm not a pre-commit user as I said. So I'd like to hear opinion from some practical pre-commit users.

AlexWaygood:
I'm a pre-commit user, and I can confirm:
- That using actionlint via pre-commit by default does not enable the shellcheck integration
- That the changes made in this PR look like they would enable the shellcheck integration by default when running actionlint via pre-commit

However, I'm not sure this is the best approach. Using `additional_dependencies: ["github.com/wasilibs/go-shellcheck/cmd/shellcheck@latest"]` means that pre-commit will always install the latest version of shellcheck, which could lead to your CI unexpectedly and unpredictably breaking if shellcheck cuts a new release and you're running actionlint in CI via pre-commit. Another approach might be to add a note to actionlint's documentation saying that users will need to explicitly list shellcheck as an additional dependency when they use actionlint as a pre-commit hook if they want the shellcheck integration. That way users are forced to consider which shellcheck version they want installed by pre-commit as part of the actionlint hook, and are able to easily pin shellcheck to a specific version.

You can take a look at the pre-commit configuration I'm adding in https://github.com/astral-sh/ruff/pull/15021/files as an example of how to use actionlint as a pre-commit hook while enabling the shellcheck integration

rhysd:
Thank you for the discussion.

Then I feel the followings are right directions:

- Create a new pre-commit hook which utilizes `go-shellcheck` in addition to `actionlint`, `actionlint-docker`, `actionlint-system` for those who prefer the go-shellcheck approach.
- Update the configuration example in the [pre-commit guide](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#pre-commit) to setup both actionlint and shellcheck as @AlexWaygood pointed.

PR #483, opened 2024-11-22
support commands specified as exe/args

### LINKED ISSUE #480: Support commands specified as space-separated exe/args
I have packaged shellcheck into a form that can be executed with `go run`

https://github.com/wasilibs/go-shellcheck

I'd like to then specify it when running actionlint so that when debugging a CI failure, devs can copy/paste to reproduce the command without any tool installation / version mismatches.

```
go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.4 -shellcheck="go run github.com/wasilibs/go-shellcheck/cmd/shellcheck@343642672fe6b725a201c27a742cbed0f5c5075a"
```

This patch seems to work well with that pattern

https://github.com/rhysd/actionlint/compare/main...anuraaga:process-args?expand=1

Is it something that could be considered for the project? If so, I'll send a PR.

anuraaga:
closes #480 

If an external command (shellcheck, pyflakes) is not found, try to parse it as a exe+args using [go-shellwords](https://github.com/mattn/go-shellwords), which can allow executing without having the binary installed i.e., `go run github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.10.0`, `pipx run pyflakes`.

### DIFF
--- a/process.go
+++ b/process.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"sync"
 
+	"github.com/mattn/go-shellwords"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 	"golang.org/x/sys/execabs"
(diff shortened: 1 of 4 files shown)


rhysd:
Thank you for the quick PR. I'll check this soon.

process.go
```diff
@@ -109,18 +110,38 @@ func (proc *concurrentProcess) wait() {
...
+	p, err := execabs.LookPath(exe)
+	if err == nil {
+		return p, nil, nil
+	}
+	// See if the command string contains args. As it is best effort, we do not
+	// handle parse errors.
+	if exeArgs, _ := shellwords.Parse(exe); len(exeArgs) > 0 {
+		// We want to return the original error if this command isn't found so
+		// do not overwrite it.
+		if p, err2 := execabs.LookPath(exeArgs[0]); err2 == nil {
```
rhysd, line 137:
```suggestion
		if p, err := execabs.LookPath(exeArgs[0]); err == nil {
```

Would you shadow the outer `err` variable? It is okay because this variable is very short-lived.

process_test.go
```diff
@@ -63,6 +63,28 @@ func TestProcessRunConcurrently(t *testing.T) {
 	}
 }
 
+func TestProcessRunWithArgs(t *testing.T) {
+	var done atomic.Bool
+	p := newConcurrentProcess(1)
+	echo := testSkipIfNoCommand(t, p, "echo hello")
```
rhysd, line 69: This test doesn't fail when `newCommandRunner` fails because it calls `t.Skip`.

Instead, we need to check if `echo` exists separately.

```go
if _, err := execabs.LookPath("echo"); err != nil {
    t.Skipf("echo command is necessary to run this test: %s", err)
}

c, err := p.newCommandRunner("echo hello", false)
if err != nil {
    t.Fatalf(`parsing "echo hello" failed: %v`, err)
}
```

rhysd (changes requested):

anuraaga:
Thanks @rhysd applied the fixes

PR #505, opened 2025-01-17
chore: update actions to read go version from `go.mod`

chenrui333:
- update actions to read go version from `go.mod`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/rhysd/actionlint
 
-go 1.18
+go 1.23.5
 
 require (
 	github.com/bmatcuk/doublestar/v4 v4.7.1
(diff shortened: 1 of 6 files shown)


rhysd:
This project's `go.mod` specifies `go 1.18`. Would you explain how the `go-version-file` in this project? I'd like to ensure that proper Go version which I want to use is used in CI.

chenrui333:
it would just simplify the go version, go version in go.mod would be the source of truth the version in the project.

rhysd:
Thanks. And I understood the version selection by the following official document. It sounds good. https://go.dev/ref/mod#go-mod-file-go

CodeQL job is broken due to this change. However I think I can try to fix it.

rhysd (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/rhysd/actionlint
 
-go 1.18
+go 1.23.5
```
rhysd, line 3: I overlooked this line. This line is not acceptable. I'll remove this change from main branch.

https://github.com/rhysd/actionlint/blob/main/docs/api.md#go-version-compatibility

PR #506, opened 2025-01-17
chore: pin workflow action sha

chenrui333:
instead of using tag, which is mutable, update the action deps to use sha


rhysd:
Thanks for your suggestion. However I'm satisfied with the tags so far.

allanlewis:
Perhaps this should be reconsidered following the [`tj-actions/changed-files` incident](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised)?

PR #507, opened 2025-01-18
Run unit test GitHub Action on an ARM processor

cclauss:
Projects are speeding up GitHub Actions by running them on an ARM processor. https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources


rhysd (approved):

rhysd:
Thanks! CI passed without any error :tada:. Merged.

PR #503, opened 2025-01-16
Support ARM64 runners

martincostello:
Add support for the new `ubuntu-24.04-arm` and `ubuntu-22.04-arm` runner labels.

[_Linux arm64 hosted runners now available for free in public repositories (Public Preview)_](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/)

PR #508, opened 2025-01-20
Update example errors in README

muzimuzhi:
Follow-up to #503.

I've checked that all other error lines in the README are up-to-date.


rhysd (approved):
I missed it. Thanks.

PR #510, opened 2025-02-04
Fix actions/*-artifact false positives for GHES

### LINKED ISSUE #509: actions/download-artifact false positives for GitHub Enterprise Server
Follow-up to #468.

In a GitHub Enterprise **Server** (GHES) environment where GitHub Connect is not enabled (i.e. actions are shipped-in box with the appliance, and not downloaded from GitHub.com) the `actions/upload-artifact@v3-node20` and `actions/download-artifact@v3-node20` branches/tags/references do not exist in the actions repositories that ship with the appliance.

Accordingly, it is not possible to use actionlint v1.7.4+ in GHES as it is impossible to upgrade from v3 to v4 as it is not supported. I assume GHES doesn't enforce the Node.js v16 restriction internally in GHES environments as they are still functional.

My suggestion for a fix would be to not issue these warnings if the `GITHUB_SERVER_URL` environment variable is not `https://github.com` and does not end with `.ghe.com` (for GitHub Enterprise Cloud) ([docs](https://docs.github.com/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables)).

martincostello:
Attempt to resolve false positives for v4 artifacts actions in GitHub Enterprise Server.

Resolves #509.

### DIFF
--- a/rule_action.go
+++ b/rule_action.go
@@ -288,7 +288,7 @@ var BrandingIcons = map[string]struct{}{
 
 // These actions are still supported in GitHub Enterprise Server as the v4 back-end infrastructure
 // that the *-artifact actions use is not available, or they depend on those v4 actions.
-var OutdatedPopularActionSpecsStillSupportedByGitHubEnterpriseServer = map[string]struct{}{
+var outdatedPopularActionSpecsStillSupportedByGitHubEnterpriseServer = map[string]struct{}{
 	"actions/deploy-pages@v2":          {},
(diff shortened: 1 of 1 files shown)


codezninja:
@rhysd is this something we could add to the next release?

martincostello:
I'm going to close this out as it's gone stale and I no longer work for the company that had the need for this, so post-merge I can't test whether it solves the original problem.

Happy for someone else to pick the changes up and run with them.

PR #528, opened 2025-03-28
add repository_name to expr_sema.go

mm-chia:
This property is undocumented in the official Github documentation, however it is perfectly valid and actionlint errors out when it's being used.

Attempts to fix https://github.com/rhysd/actionlint/issues/529


rhysd:
This is not listed because the `repository_name` prop is undocumented. Do you have any pointer which describes that this is an official property?

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

rhysd:
Closing this because #529 was closed.

PR #531, opened 2025-04-11
add `models` as allowed permission

muzimuzhi:
See related GitHub blog [1] and docs [2].

[1]: https://github.blog/changelog/2025-03-18-as-of-march-29-fine-grained-pats-and-github-apps-need-updates-to-access-github-models-playground/
[2]: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions

One concern: according to [2], `models` only supports values `read` and `none`, but not `write`. Similarly `id-token` doesn't support value `read`.


bhoslepu:
@rhysd @muzimuzhi Request to you to merge this PR? At-least new addition of models permissions will not break the lint checks in workflows?

rhysd (approved):

PR #541, opened 2025-05-17
fix case insensitive object index access `github['REF']` / preserve case of env.TEST in parser

ChristopherHX:
* preserve property case of objderef to allow case sensitive env access in act

In my workflow script interpreter using your parser I faced the problem that
- env.TEST get parsed as env.test
- However on linux/macOS is env.TEST != env.test (the only context with an exception)
- Now our external interpreter has no chance to get the original case of the parsed expression

This PR preserves the original value much longer, e.g. error messages now quote the original case while case insensitivity is preserved.

I also noticed an inconsistency here e.g. `github['REF']` was marked as invalid, while `github.REF` passed that I fixed here as well.

_I am a user of actionlint as a library, not of the cli tool itself_


ChristopherHX:
Considering to remove the dependency by rewriting the whole expression lexer and parser in  <https://github.com/actions-oss/act-cli/issues/131>

PR #546, opened 2025-06-10
docs: enlighten Go newcomers about GOPATH

lkraav:


fredrikaverpil:
You can also add `actions/setup-go@v5`, which will add the path to `~/go/bin`:

```yml
name: "lint"
on:
  push:
    branches:
      - master
  pull_request:
  workflow_dispatch:
jobs:
  action-lint:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v4
      - uses: actions/setup-go@v5
        with:
          go-version: 'stable'
      - name: Install actionlint
        run: go install github.com/rhysd/actionlint/cmd/actionlint@latest
      - name: Run actionlint
        run: actionlint
```

rhysd:
I don't think this is necessary. Just googling "how to install go" is enough. We don't need to duplicate it here. And this instruction has some issues:

- GOPATH is not always `~/go`
- It is not valid for Windows

PR #549, opened 2025-07-02
Add support for ppc64le/s390x runners

bluca:
Projects can request IBM to assign ppc64le/s390x GHA runners. The linter doesn't know about these images, so it fails when they are used.

https://github.com/IBM/actionspz/blob/main/docs/onboarding.md


rhysd:
These are not official runners. Please use self-hosted-runner config.

https://github.com/rhysd/actionlint/blob/main/docs/config.md

PR #557, opened 2025-08-05
Add YAML anchor and alias support for GitHub Actions workflows

### LINKED ISSUE #133: False Positive On Aliases In Mapping Nodes
Actionlint complains `"with" section is alias node but mapping node is expected [syntax-check]` on valid YAML syntax when anchors and aliases are used with mapping nodes:

```yaml
- name: Run my-action.
  run: my-organization/my-action
  with: &my-action-inputs
    a: b
    c: d
    e: f
- name: Run my-action again.
  run: my-organization/my-action
  with: *my-action-inputs
```

I am uncertain how one would suppress the error.

alexaandru:
Implements two-pass YAML parsing to resolve anchors (&anchor) and aliases (*alias) before linting, matching GitHub Actions' native YAML support.

- Add containsAnchorsOrAliases() to detect anchor usage
- Add resolveYAMLAnchors() using unmarshal/marshal strategy
- Modify Parse() to resolve anchors when present
- Add comprehensive test suite covering various anchor scenarios
- Preserve backward compatibility for workflows without anchors
- Maintain error detection and line number reporting after resolution

Fixes workflows that previously failed with "alias node but mapping node is expected" errors when using YAML anchors supported by GitHub Actions.

Fixes #133 

The support for this just landed: https://github.com/actions/runner/issues/1182


alexaandru:
Please note that both lint and test FAIL but they also FAIL on `main` branch. This PR didn't introduce any new test failures, they are the same as on `main`. It did however introduce plenty of tests for the feature and they are all passing.

neodc:
Any idea of when this could be merged and released?

rhysd:
It seems that #557 and #568 has the same goal and different implementations. I need to check both but I don't have enough time for now. Let me check this PR after releasing the next version.

rhysd:
I'm closing and reopening this PR to kick CI workflows because GitHub seemed to dismiss 'Approve CI' button.

vivodi:
Can you fix the failing checks? @alexaandru

rhysd:
Thank you for working on this feature and the approach used in this branch is interesting.

However I found a critical issue that the error position can be completely corrupted and cannot be recovered.

For example

```yaml
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: env | grep MY_TEST_
        env: &default_env
          INVALID_ENV: []
      - run: env | grep MY_TEST_
        env: *default_env
```

This workflow causes the following errors:

```
test.yaml:6:30: expected scalar node for string value but found sequence node with "!!seq" tag [syntax-check]
test.yaml:9:30: expected scalar node for string value but found sequence node with "!!seq" tag [syntax-check]
```

However the positions `6:30` and `9:30` are actually not indicating the error positions in the original workflow because we parse a converted workflow generated by go-yaml. This issue is critical because users are not able to understand where the errors occurred.

Issue #559, opened 2025-08-06
GitHub Actions Node 24 Support


ncalteen:
GitHub Actions now supports Node v24. Would it be possible to add `node24` as an allowed value for the `runs.using` property?

Infinoid:
We're already seeing this new value in the wild: the popular `actions/checkout` action uses it in [their recent v5.0.0 release](https://github.com/actions/checkout/releases/tag/v5.0.0).

PR #561, opened 2025-08-08
Allow node 24 for actions

salmanmkc:
Closes https://github.com/rhysd/actionlint/issues/559
Runner now has node24 support, and this linter is causing CI to fail since it doesn't allow node24.


chenrui333 (approved):

chenrui333:
cc @rhysd

rhysd (approved):

PR #568, opened 2025-08-28
feat: Support YAML anchors and aliases in parser

srz-zumix:
This change adds support for YAML anchors and aliases in the workflow parser.
https://github.com/actions/runner/issues/1182#issuecomment-3150797791

A new helper function `resolveAlias` is introduced to dereference alias nodes before they are processed by the parsing functions. This function is called at the beginning of each parsing function to ensure that the parser is always working with the actual value nodes.

A new test case is added to verify that YAML anchors are correctly parsed. An existing test is also updated to ignore the `version` field in the SARIF output to make the tests more stable.

### DIFF
--- a/linter_test.go
+++ b/linter_test.go
@@ -16,6 +16,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"golang.org/x/sys/execabs"
 )
 
@@ -459,7 +460,15 @@ func TestLinterFormatErrorMessageInSARIF(t *testing.T) {
 		panic(err)
 	}
 
-	if diff := cmp.Diff(want, have); diff != "" {
+	if diff := cmp.Diff(
+		want,
+		have,
+		cmpopts.IgnoreMapEntries(func(k string, v interface{}) bool {
+			// Built version is not stable across Go versions and environments.
+			// To avoid this instability, ignore "version" field in SARIF.
+			return k == "version"
+		}),
+	); diff != "" {
 		t.Fatal(diff)
 	}
 }
--- a/parse.go
+++ b/parse.go
@@ -65,6 +65,13 @@ func (p *parser) errorfAt(pos *Pos, format string, args ...interface{}) {
 	p.errorAt(pos, m)
 }
 
+func (p *parser) resolveAlias(n *yaml.Node) *yaml.Node {
+	for n.Kind == yaml.AliasNode {
+		n = n.Alias
+	}
+	return n
+}
+
 func (p *parser) errorf(n *yaml.Node, format string, args ...interface{}) {
 	m := fmt.Sprintf(format, args...)
 	p.error(n, m)
@@ -84,6 +91,7 @@ func (p *parser) unexpectedKey(s *String, sec string, expected []string) {
 }
 
 func (p *parser) checkNotEmpty(sec string, len int, n *yaml.Node) bool {
+	n = p.resolveAlias(n)
 	if len == 0 {
 		p.errorf(n, "%q section should not be empty", sec)
 		return false
@@ -92,6 +100,7 @@ func (p *parser) checkNotEmpty(sec string, len int, n *yaml.Node) bool {
 }
 
 func (p *parser) checkSequence(sec string, n *yaml.Node, allowEmpty bool) bool {
+	n = p.resolveAlias(n)
 	if n.Kind != yaml.SequenceNode {
 		p.errorf(n, "%q section must be sequence node but got %s node with %q tag", sec, nodeKindName(n.Kind), n.Tag)
 		return false
@@ -100,6 +109,7 @@ func (p *parser) checkSequence(sec string, n *yaml.Node, allowEmpty bool) bool {
 }
 
 func (p *parser) checkString(n *yaml.Node, allowEmpty bool) bool {
+	n = p.resolveAlias(n)
 	// Do not check n.Tag is !!str because we don't need to check the node is string strictly.
 	// In almost all cases, other nodes (like 42) are handled as string with its string representation.
 	if n.Kind != yaml.ScalarNode {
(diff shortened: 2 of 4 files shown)


muzimuzhi:
A previously opened PR #557 implements the same feature. Would you like to compare your implementations and cooperation with the author of #557?

srz-zumix:
#557 involves expanding the anchor/alias beforehand and then using it to proceed with the subsequent process as is. I think it's a good idea.
This approach, on the other hand, is a simple implementation that directly returns the destination of the alias if the target node is an alias. It requires adding processing to each function.
Additionally, the issues with the tests have already been resolved.
https://github.com/srz-zumix/actionlint/actions/runs/17301408326?pr=1

Regardless of the implementation method, I sincerely hope that actionlint will support anchor/alias.

vivodi:
@rhysd Any idea of when this could be merged and released?

rhysd:
It seems that #557 and #568 has the same goal and different implementations. I need to check both but I don't have enough time for now. Let me check this PR after releasing the next version.

rhysd:
One quick question: What happens if two anchors are mutually recursively referencing each other? I'd appreciate if such edge case is tested. And I would also appreciate error test cases (e.g. undefined anchor).

ChristopherHX:
<Details>
<Summary>I was wrong, missed an other cyclic reference</Summary>

> One quick question: What happens if two anchors are mutually recursively referencing each other?

I had expected issues like this, but this is a false positive yaml v3 for go errors out in this case.

Here a test case to verify:

https://github.com/actions-oss/act-cli/blob/67f0e3d8520afc84234212e8af74825c0d939207/pkg/model/anchors_test.go#L10

The error is anchor b not found, while the parser is at ref: *b.

```go
func TestVerifyCycleIsInvalid(t *testing.T) {
	var node yaml.Node
	err := yaml.Unmarshal([]byte(`
a: &a
  ref: *b

b: &b
  ref: *a
`), &node)
	assert.Error(t, err)
}
```

I suggest to copy this test into your codebase, just in case the abandoned yaml lib is going to be replaced / changes behavior. Maybe better use actionlint for reading it here.

After this I stripped my own cyclic prevention code, because it could never get code coverage.

https://go.dev/play/p/eDfPhWtTf3a

---
I decided to explode all yaml anchors in my projects, otherwise too many code changes and the possibility new anchor unsupported code is added without notice.

</Details>

srz-zumix:
I have expanded our test coverage to include scenarios involving recursion.
https://github.com/srz-zumix/actionlint/pull/2

rhysd:
@ChristopherHX Thank you for exploring the edge case. That's definitely helpful. I'll consider it.

vivodi:
Can you fix the failing checks? @srz-zumix

rhysd:
I tried this branch today but I immediately noticed it didn't support `<<:`. Any ideas?

```yaml
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - &default_step
        run: echo hello
        shell: bash
      - <<: *default_step
        run: echo hello
```

ChristopherHX:
GitHub never implemented support for `<<` in Actions, having actionlint supporting this could causes confusion.

I do not know what this PR shows in this case if you use it anyway.

Some people in the GitHub Actions FR channel were not really happy about this detail.

rhysd:
Okay, thanks for the info. I'll try some examples on GitHub Actions to ensure it.

> I do not know what this PR shows in this case if you use it anyway.

It actually just says 'unknown key "<<:"' because go-yaml doesn't handle special key `<<` at all when parsing inputs into `yaml.Node` tree.

rhysd:
I confirmed GitHub Actions runtime also does not handle `<<` merge key. It handles `<<` as normal YAML key. So we don't need to care about it at this point (though some better error message would be great).

rhysd:
Another problem I found in this branch was duplicate errors.

```yaml
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: env
        env: &default_env
          INVALID_ENV: []
      - run: env
        env: *default_env
```

causes

```
test.yaml:9:24: expected scalar node for string value but found sequence node with "!!seq" tag [syntax-check]
  |
9 |           INVALID_ENV: []
  |                        ^~
test.yaml:9:24: expected scalar node for string value but found sequence node with "!!seq" tag [syntax-check]
  |
9 |           INVALID_ENV: []
  |                        ^~
```

There is no way to understand where the second error was actually caused for users.

rhysd:
Anyway, the implementation basically looks good to me. I will supersede this work.

rhysd:
(I noticed that `<<` key is only in the old specification YAML 1.1. So it would never be implemented in GitHub Actions in the future.)

rhysd:
Note:

This implementation does not consider infinite recursive references like

```yaml
- run: env
  env: &recursive
    FOO: *recursive
```

Simply trying to resolve it can cause infinite recursive calls.

rhysd:
Note:

This implementation also needs to consider what we should do when same anchor is defined multiple times.

```yaml
- run: ...
  env: &env
- run: ...
  # Define again
  env: &env
```

- [YAML 1.2](https://yaml.org/spec/1.2.2/#71-alias-nodes) defines that the later one overrides the former one.
- From linter point of view this can be typically a bug of the workflow. So complaining about overriding anchors may be a good idea.
  ```yaml
  - run: ...
    # Oops mistook * for & (`env: *env` was correct)
    env: &env
  ```

rhysd:
Although this PR looks unmerged, the patch in this branch was committed to the main branch at 8ed4c46f9a0fadb046903c2909901e82e16e5c2c.

lkraav:
> Although this PR looks unmerged, the patch in this branch was committed to the main branch at [8ed4c46](https://github.com/rhysd/actionlint/commit/8ed4c46f9a0fadb046903c2909901e82e16e5c2c).

Superb :raised_hands: Can it be tagged into a release and let the world enjoy?

rhysd:
1.7.10 was released: https://github.com/rhysd/actionlint/releases/tag/v1.7.10

PR #577, opened 2025-10-03
Fix type of `job.check_run_id` property

muzimuzhi:
`job.check_run_id` property has type `number`, not `string`.

- GitHub Docs https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#job-context
- example run
  https://github.com/dhimmel/dump-actions-context/actions/runs/18170960955/job/51725241954#step:12:354

I asked why it has `number` type [here](https://github.com/orgs/community/discussions/8945#discussioncomment-14481081), but got no response from GitHub staff.


rhysd (approved):

rhysd:
Ah, thanks!

PR #602, opened 2025-12-22
fix: add artifact-metadata

martincostello:
Add support for the `artifact-metadata` permission.

[Docs](https://docs.github.com/actions/reference/workflows-and-actions/workflow-syntax#defining-access-for-the-github_token-scopes).


rhysd:
Thank you for catching this! It seemed to be added 1 month ago.

https://github.com/github/docs/commit/61c63d5a5f250f383839d86fe7c3ae5e0a35ba3e

PR #603, opened 2025-12-30
fix: Globs starting with './' never match

### LINKED ISSUE #521: Globs starting with `./` never match
This job will never run, because globs starting with `./` are never matched:

```yaml
on:
  pull_request:
    paths:
      - './src/**'
```

This behavior is surprising and easy to check. Let's add a check for it.

majiayu000:
Fixes #521

## Changes
- Add validation check in `ValidatePathGlob` to detect and warn about './' prefix in path globs
- The error message explains that './' prefix never matches and suggests removing it


glob.go
```diff
@@ -251,5 +251,10 @@ func ValidatePathGlob(pat string) []InvalidGlobPattern {
 			{"path value must not end with spaces", len(pat)},
 		}
 	}
+	if strings.HasPrefix(pat, "./") {
+		return []InvalidGlobPattern{
+			{"path filter \"./\" prefix never matches any file paths. remove the \"./\" prefix from the path filter", 0},
+		}
+	}
```
rhysd: This implementation is incomplete. For example `!./foo` is not detected.

rhysd (changes requested):

Issue #612, opened 2026-01-27
Undefined function "case"


ysbaddaden:
The action complains that the `case` function is undefined, but it does exist in practice:

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case

muzimuzhi:
Support for `case` function is added recently:

- actions/runner@6510776 (Add support for case function (#<span></span>4147), 2025-12-17)
- github/docs@434787e (Add case function documentation (#<span></span>58868), 2026-01-26)

rhysd:
Fixed by #614. Thanks @heppu for the fix.

heppu:
Any plans for making a release which would include this? @rhysd

rhysd:
I'm thinking about this weekend.

rhysd:
v1.711 was released: https://github.com/rhysd/actionlint/releases/tag/v1.7.11

PR #614, opened 2026-02-03
Support case func

heppu:
Add support for #612


muzimuzhi:
`docs/checks.md` contains another example of "undefined function" error message, which lists all the available functions thus needs to be updated too.

heppu:
@muzimuzhi updated docs

PR #615, opened 2026-02-05
Add windows-2025-vs2026 and macos-26-large runners

hugovk:
New today: https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/

> To use the the new runner image, update the runs-on: target in your YAML workflow file to `windows-2025-vs2026`. For additional information on the available windows runner images, visit [our doucmentation](https://docs.github.com/actions/reference/runners/github-hosted-runners?versionId=free-pro-team%40latest&productId=actions&restPage=concepts%2Crunners%2Cgithub-hosted-runners).

> We’ve released the GitHub Actions macOS 26 Intel runner image in public preview for larger runners, providing the latest macOS and Xcode tooling for Intel-based workflows. To use it, update the runs-on: target in your YAML workflow file to `macos-26-large`. For complete details on available software and versions, visit the [macOS runner images documentation](https://docs.github.com/actions/reference/runners/larger-runners#available-macos-larger-runners-and-labels).

PR #572, opened 2025-09-21
Add new macOS runner labels and remove dropped `windows-2019` label

muzimuzhi:
- add `macos-26` and `macos-26-xlarge`
  - https://github.blog/changelog/2025-09-11-actions-macos-26-image-now-in-public-preview/
- remove `windows-2019` which was dropped in Jun 2025
  - https://github.blog/changelog/2025-04-15-upcoming-breaking-changes-and-releases-for-github-actions/#windows-server-2019-is-closing-down
  - https://github.com/actions/runner-images/issues/12045
- add `macos-15-intel`
  - https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/#what-you-need-to-do
  - https://github.com/actions/runner-images/issues/13045

Fixes #570.
Fixes #571.

PR #616, opened 2026-02-05
Fix a runner name typo in Changelog

muzimuzhi:
#572 added support for the `macos-26-xlarge` runner label, but `macos-26-large` (without `x`) was recorded in `CHANGELOG.md`.

Noticed when reviewing #615, which adds support for the indeed `macos-26-large` label.


rhysd:
Thanks. I applied the same fix to the GitHub release page (because the changelog file is generated from the GitHub release notes).

PR #629, opened 2026-03-02
Add `macos-26-intel`

hugovk:
https://github.blog/changelog/2026-02-26-macos-26-is-now-generally-available-for-github-hosted-runners/

The other three are already here.

Also bump the CI to use `macos-26-intel` instead of `macos-15-intel`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,7 @@ jobs:
   unit-tests:
     name: Unit tests
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


muzimuzhi:
I doubt that the `macos-26-intel` label was initially listed when the blog was posted.

It was missing from
- https://github.com/actions/runner-images/issues/13637

but suddenly appeared in
- https://github.com/actions/runner-images/issues/13739

Anyway, thank you for catching this.

rhysd:
Now it is listed in the official document so I'm sure it's available. Thanks.

https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

PR #649, opened 2026-04-15
add config-secrets option to allow-list secrets in workflows

### LINKED ISSUE #609: Add a `secrets` configuration option and check much like `config-variables`
Hello,

I've been using actionlint which is a great tool and I particularly like the fact that we can list the variables available at the repo level within the `config-variables` array in `.github/actionlint.yml`. I think it would be great to have the same option to create an allow-list of `secrets` for repository secrets.

tdeo:
Fixes https://github.com/rhysd/actionlint/issues/609

I did use claude-code to open this because I'm not familiar at all with go. I tried to review the changes as best I could but apologize in advance if this is not good enough.

## Summary

- Adds a new `config-secrets` configuration option in `actionlint.yaml`, parallel to the existing `config-variables` option
- When set to an array of secret names, actionlint reports any `secrets.XXXX` access where XXXX is not in the list (case-insensitive)
- Default value `null` disables the check; an empty array disallows all secrets
- Updates `docs/config.md` with documentation and usage example

## Test plan

- [ ] New project-level integration test in `testdata/projects/config_secrets/`
- [ ] New unit tests in `TestExprSemanticsCheckOK` for `secret`, `known secret`, and `secret name is case insensitive`
- [ ] New unit tests in `TestExprSemanticsCheckError` for `no secret is allowed` and `unknown secret`
- [ ] All existing tests pass (`go test ./...`)
- [ ] `staticcheck ./...` passes cleanly

### DIFF
--- a/expr_sema.go
+++ b/expr_sema.go
@@ -374,17 +374,19 @@ type ExprSemanticsChecker struct {
 	availableContexts     []string
 	availableSpecialFuncs []string
 	configVars            []string
+	configSecrets         []string
 }
 
 // NewExprSemanticsChecker creates new ExprSemanticsChecker instance. When checkUntrustedInput is
 // set to true, the checker will make use of possibly untrusted inputs error.
-func NewExprSemanticsChecker(checkUntrustedInput bool, configVars []string) *ExprSemanticsChecker {
+func NewExprSemanticsChecker(checkUntrustedInput bool, configVars []string, configSecrets []string) *ExprSemanticsChecker {
 	c := &ExprSemanticsChecker{
 		funcs:           BuiltinFuncSignatures,
 		vars:            BuiltinGlobalVariableTypes,
 		varsCopied:      false,
 		githubVarCopied: false,
 		configVars:      configVars,
+		configSecrets:   configSecrets,
 	}
 	if checkUntrustedInput {
 		c.untrusted = NewUntrustedInputChecker(BuiltinUntrustedInputs)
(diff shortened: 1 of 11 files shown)


rhysd:
Since this change seems to be generated by Claude, let me ask a question. Did you review the generated code, confirmed the test results, and validated the behavior with actual config files?

expr_sema.go
```diff
@@ -713,6 +718,33 @@ func (sema *ExprSemanticsChecker) checkConfigVariables(n *ObjectDerefNode) {
...
+		return
+	}
+	if len(sema.configSecrets) == 0 {
+		sema.errorf(
+			n,
+			"no secret is allowed since the secrets list is empty in actionlint.yaml. you may forget adding the secret %q to the list",
+			n.Property,
+		)
+		return
+	}
```
rhysd, line 732: I don't think this fast path is necessary because the below error message is comprehensive enough.

rhysd (changes requested):

tdeo:
> Since this change seems to be generated by Claude, let me ask a question. Did you review the generated code, confirmed the test results, and validated the behavior with actual config files?

Hello, that's a very good, while I didn't know anything about go and not much about this project's internal, what I did review myself was:
- that the generated files under `testdata/` made sense, that they were implemented in a very similar way to what exists for config variables and that the output is the behavior I would expect for a project with that workflow and configuration
- that the implementation of `checkConfigSecrets` seemed to be coherent with what existed for `checkConfigVariables`
- that the test suite still worked locally (claude was able to run it by itself)
- I did see that the signature change to `NewExprSemanticsChecker` may not be what we want long term and possibly it would be better for it to receive a config object instead of individual values but I didn't think such a change was appropriate in this PR and I had no idea how to start

Overall, I did think opening a PR was a good way to get the conversation started about https://github.com/rhysd/actionlint/issues/609

expr_sema.go
tdeo: Do you think we should also remove the same fast path from `checkConfigVariables` just above?

tdeo: @rhysd do you have any opinion whether I should handle checkConfigVariables at the same time?

PR #651, opened 2026-04-17
fix deadlock in shellcheck integration on darwin by using cmd.Stdin

### LINKED ISSUE #650: Deadlock in shellcheck integration on macOS: process.go:32 blocks writing to unstarted child stdin
### Summary
actionlint hangs (spinning ~450% CPU, never exits) on workflows that have **≥2 `run:` steps with ~1 KB+ shell scripts** when running on **darwin/arm64**. All shellcheck worker goroutines deadlock at `process.go:32` writing to the shellcheck child's stdin pipe.

### Reproduction

macOS arm64 (Darwin 25.4.0, M-series), shellcheck 0.11.0.

```yaml
# save as deadlock.yaml
name: N50
on: push
jobs:
  j:
    runs-on: ubuntu-latest
    steps:
      - run: |
          echo hello
          echo hello
          # … (50 identical `echo hello` lines)
      - run: |
          echo hello
          echo hello
          # … (50 identical `echo hello` lines)
```

```
$ actionlint deadlock.yaml
# hangs forever, ~450% CPU
```

Also reproduces on 1.7.10, 1.7.11, and 1.7.12 official darwin_arm64 release binaries (Go 1.25.5, 1.25.7, and 1.26.1 respectively) — so it's not a Go-version regression.

Running with `-shellcheck=` (integration disabled) finishes in ~20 ms. Running shellcheck standalone with the exact args actionlint passes (`--norc -f json -x --shell bash -e SC1091,SC2194,SC2050,SC2153,SC2154,SC2157,SC2043 -`) on the extracted scripts finishes in <100 ms, including 5 in parallel. So the hang is in actionlint, not shellcheck.

### Diagnosis

`SIGQUIT` stack trace (abridged) — every shellcheck worker goroutine is blocked here:

```
goroutine N [IO wait]:
internal/poll.(*pollDesc).waitWrite(...)
internal/poll.(*FD).Write
os.(*File).Write
io.WriteString
github.com/rhysd/actionlint.(*cmdExecution).run
    actionlint/process.go:32 +0x114
github.com/rhysd/actionlint.(*concurrentProcess).run.func1
    actionlint/process.go:99
golang.org/x/sync/errgroup.(*Group).Go.func1
```

The offending pattern in `process.go`:

```go
func (e *cmdExecution) run() ([]byte, error) {
    cmd := exec.Command(e.cmd, e.args...)
    cmd.Stderr = nil
    p, err := cmd.StdinPipe()            // pipe created, child NOT started
    ...
    if _, err := io.WriteString(p, e.stdin); err != nil { ... }   // ← blocks here
    p.Close()
    ...
    stdout, err = cmd.Output()           // cmd.Start() happens only here
}
```

Writing to `StdinPipe()` **before** calling `cmd.Start()` relies on the kernel pipe buffer to absorb the whole script. On darwin/arm64 the usable buffer against an unstarted child appears to be small enough (empirically <1 KB per pipe under contention) that any moderately sized script deadlocks. Once N≥2 workers all block pre-start, no reader ever attaches.

### Suggested fix

Let Go's stdlib handle stdin copying after `Start()`:

```go
func (e *cmdExecution) run() ([]byte, error) {
    cmd := exec.Command(e.cmd, e.args...)
    cmd.Stdin = strings.NewReader(e.stdin)    // stdlib spawns copier after Start
    if e.combineOutput {
        return cmd.CombinedOutput()
    }
    return cmd.Output()
}
```

`cmd.Output()` / `cmd.CombinedOutput()` internally start a goroutine to copy from `cmd.Stdin` to the child's stdin *after* `cmd.Start()`, which eliminates the pre-start write and is the canonical pattern in the Go stdlib docs.

### Environment

- actionlint 1.7.12 (also 1.7.11, 1.7.10) darwin_arm64
- shellcheck 0.11.0
- macOS Darwin 25.4.0, M-series Apple silicon
- Go 1.25.5 / 1.25.7 / 1.26.1 (per the release binary)

### Workaround

Pass `-shellcheck=` on the CLI, or run actionlint inside a Linux container (larger pipe buffers mask the bug).

attehuhtakangas:
Fixes #650.

`cmdExecution.run` used to create a stdin pipe with `cmd.StdinPipe()`, write the full script to it, close the writer, and only then call `cmd.Output()` which internally calls `cmd.Start()`. That relies on the kernel pipe buffer being large enough to absorb the whole payload before a reader is attached. On darwin, under concurrent load, the buffer fills up and every worker goroutine ends up blocked on `internal/poll.(*FD).Write -> waitWrite` inside `io.WriteString`, so no child ever gets started and actionlint spins until killed.

This replaces the manual pipe management with `cmd.Stdin = strings.NewReader(e.stdin)`. `cmd.Output()` / `cmd.CombinedOutput()` then spawn a copy goroutine *after* `Start()`, which is the canonical os/exec pattern.

The bug reproduces reliably on `darwin_arm64` (tested on 1.7.10, 1.7.11, and 1.7.12 official release binaries) with any workflow that has ≥2 `run:` steps whose scripts combined cross the pipe-buffer threshold — the smallest case I could trim it to was a 2 KB workflow with 50 `echo hello` lines per step.

## Changes

- `process.go` — replace the pre-`Start()` pipe write with `cmd.Stdin = strings.NewReader(...)`. No behavior change on Linux, fixes the deadlock on darwin.
- `process_test.go` — add `TestProcessConcurrentStdinDoesNotDeadlock`, which runs 5 concurrent `cat` invocations with a 64 KiB stdin payload guarded by a 10 s watchdog. Confirmed:
  - FAILS on `main` (watchdog fires at 10 s)
  - PASSES on this branch

## Test evidence

Before the fix, running on my local `darwin/arm64` M-series against the issue's 50-echo repro:

```
$ actionlint deadlock.yaml   # spins at ~450% CPU, never exits
```

After the fix (same binary built from this branch):

```
$ ./actionlint deadlock.yaml
$ echo $?
0
```

Full test suite (`go test ./...`) still passes.

### DIFF
--- a/process.go
+++ b/process.go
@@ -3,8 +3,8 @@ package actionlint
 import (
 	"context"
 	"fmt"
-	"io"
 	"os/exec"
+	"strings"
 	"sync"
 
 	"github.com/mattn/go-shellwords"
@@ -24,18 +24,14 @@ type cmdExecution struct {
 func (e *cmdExecution) run() ([]byte, error) {
 	cmd := exec.Command(e.cmd, e.args...)
 	cmd.Stderr = nil
-
-	p, err := cmd.StdinPipe()
-	if err != nil {
-		return nil, fmt.Errorf("could not make stdin pipe for %s process: %w", e.cmd, err)
-	}
-	if _, err := io.WriteString(p, e.stdin); err != nil {
-		p.Close()
-		return nil, fmt.Errorf("could not write to stdin of %s process: %w", e.cmd, err)
-	}
-	p.Close()
+	// Set stdin via an io.Reader so that exec.Cmd pipes the bytes to the child
+	// after Start(). Writing to cmd.StdinPipe() before Start() relies on the
+	// kernel pipe buffer being large enough to absorb the whole payload, which
+	// deadlocks on darwin when multiple workers run concurrently (issue #650).
+	cmd.Stdin = strings.NewReader(e.stdin)
 
 	var stdout []byte
+	var err error
 	if e.combineOutput {
 		stdout, err = cmd.CombinedOutput()
 	} else {
--- a/process_test.go
+++ b/process_test.go
@@ -176,6 +176,49 @@ func TestProcessInputStdin(t *testing.T) {
 	}
 }
 
+// Regression test for issue #650: concurrent runs with a stdin payload larger
+// than the kernel pipe buffer used to deadlock on darwin because the payload
+// was written to cmd.StdinPipe() before cmd.Start().
+func TestProcessConcurrentStdinDoesNotDeadlock(t *testing.T) {
+	p := newConcurrentProcess(5)
+
+	// 64 KiB is above the default pipe buffer size on darwin and Linux so it
+	// forces the stdin copy to happen after the child has started.
+	payload := strings.Repeat("x", 64*1024)
+
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		cmds := make([]*externalCommand, 0, 5)
+		for i := 0; i < 5; i++ {
+			cat := testSkipIfNoCommand(t, p, "cat")
+			cat.run(nil, payload, func(b []byte, err error) error {
+				if err != nil {
+					t.Errorf("cat failed: %v", err)
+					return err
+				}
+				if len(b) != len(payload) {
+					t.Errorf("cat output length %d, want %d", len(b), len(payload))
+				}
+				return nil
+			})
+			cmds = append(cmds, cat)
+		}
+		for _, c := range cmds {
+			if err := c.wait(); err != nil {
+				t.Errorf("cat wait failed: %v", err)
+			}
+		}
+		p.wait()
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(10 * time.Second):
+		t.Fatal("concurrent stdin writes deadlocked — see issue #650")
+	}
+}
+
 func TestProcessErrorCommandNotFound(t *testing.T) {
 	p := newConcurrentProcess(1)
 	c := &externalCommand{


rhysd (approved):
I could not reproduce the test failure by just copying&pasting the test case on my old Intel Mac. However this change looks more clean than previous and the CI passed. Merging.

attehuhtakangas:
@rhysd thank you for merging the fix! 🙏 Any chance to get this released soon?

ibetitsmike:
@rhysd any chance for a release? hitting this heavily now

kjanat:
> I could not reproduce the test failure by just copying&pasting the test case on my old Intel Mac. However this change looks more clean than previous and the CI passed. Merging.

Regarding the review note about not reproducing this on an Intel Mac: I made a standalone reproducer that avoids depending on the timing of the concurrent regression test. It generates a synthetic `run:` block and bisects the smallest payload where actionlint stops returning.

Happy to see it's fixed already on main! Would appreciate a new release.

On affected v1.7.12/Linux, it found:

```text
65518 B  ok
65519 B  HANG

with actionlint prefix: 65536 bytes
fresh pipe capacity:    65536 bytes (exact match)
```

Against the current development build containing #651, the same script completed successfully with a 131072-byte block despite the 65536-byte pipe capacity.

Run it with:

```sh
uv run repro.py
# or just
python repro.py
```

[repro.py](https://github.com/user-attachments/files/30831269/repro.py)

Pipe-capacity reporting uses Linux `F_GETPIPE_SZ`, but the reproduction and bisection still run on macOS and Windows.

<details><summary><code>repro.py</code> contents</summary>

```python
#!/usr/bin/env -S uv run --script
# requires-python = ">=3.14"
# dependencies = []
# ///
"""Reproduce actionlint's pre-Start stdin deadlock and bisect its threshold.

Requirements:
    - ``actionlint`` and ``shellcheck`` must be installed and available on PATH.
    - Run against an affected actionlint release, such as v1.7.12, to reproduce
      the deadlock. Builds containing PR #651 should complete without hanging.

Run on Linux, macOS, or Windows:
    uv run repro.py

Use explicit executable paths when they are not on PATH:
    uv run repro.py --actionlint ACTIONLINT --shellcheck SHELLCHECK

Raise the largest tested run block if the platform has a larger pipe capacity:
    uv run repro.py --ceiling 1048576

Change how long a non-returning actionlint invocation must run before it counts
as hung:
    uv run repro.py --timeout 10

The script exits with status 2 if a dependency is missing or actionlint returns
an unexpected nonzero status. A successful run either prints the smallest
hanging run-block size or reports that no deadlock occurred below ``--ceiling``.
Linux additionally reports the capacity of a freshly created pipe; capacity
reporting is unavailable on platforms without ``F_GETPIPE_SZ``, but the
reproduction and bisection still run.
"""

import argparse
import os
import platform
import shutil
import subprocess
import sys
import tempfile
from enum import Enum, auto
from pathlib import Path

ACTIONLINT_PREFIX = b"set -eo pipefail\n"
MINIMUM_SCRIPT_BYTES = 2


class ProbeResult(Enum):
    COMPLETED = auto()
    HUNG = auto()


class ProbeFailure(RuntimeError):
    """Raised when actionlint exits instead of completing or timing out."""


class Arguments(argparse.Namespace):
    actionlint: str = "actionlint"
    shellcheck: str = "shellcheck"
    timeout: float = 6.0
    ceiling: int = 131_072


def positive_int(value: str) -> int:
    parsed = int(value)
    if parsed <= 0:
        raise argparse.ArgumentTypeError("must be greater than zero")
    return parsed


def positive_float(value: str) -> float:
    parsed = float(value)
    if parsed <= 0:
        raise argparse.ArgumentTypeError("must be greater than zero")
    return parsed


def parse_args() -> Arguments:
    parser = argparse.ArgumentParser(
        description=(
            "Generate synthetic workflows and find the smallest run block that "
            "makes an affected actionlint build deadlock before starting shellcheck."
        )
    )
    _ = parser.add_argument(
        "--actionlint",
        default="actionlint",
        help="actionlint executable or command name (default: %(default)s)",
    )
    _ = parser.add_argument(
        "--shellcheck",
        default="shellcheck",
        help="shellcheck executable or command name (default: %(default)s)",
    )
    _ = parser.add_argument(
        "--timeout",
        type=positive_float,
        default=6.0,
        help="seconds before a probe counts as hung (default: %(default)s)",
    )
    _ = parser.add_argument(
        "--ceiling",
        type=positive_int,
        default=131_072,
        help="largest run block to probe, in bytes (default: %(default)s)",
    )
    arguments = Arguments()
    _ = parser.parse_args(namespace=arguments)
    return arguments


def resolve_executable(value: str) -> str:
    executable = shutil.which(value)
    if executable is None:
        raise ProbeFailure(f"executable not found: {value}")
    return executable


def version_output(executable: str, flag: str) -> str:
    process = subprocess.run(
        [executable, flag],
        check=False,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        text=True,
    )
    output = process.stdout.strip()
    if process.returncode != 0:
        raise ProbeFailure(
            f"{executable} {flag} exited with status {process.returncode}:\n{output}"
        )
    return output


def fresh_pipe_capacity() -> int | None:
    try:
        import fcntl
    except ImportError:  # pragma: no cover - unavailable on Windows
        return None

    if not hasattr(fcntl, "F_GETPIPE_SZ"):
        return None

    read_fd, write_fd = os.pipe()
    try:
        return int(fcntl.fcntl(write_fd, fcntl.F_GETPIPE_SZ))
    finally:
        os.close(read_fd)
        os.close(write_fd)


def linux_pipe_soft_limit() -> int | None:
    path = Path("/proc/sys/fs/pipe-user-pages-soft")
    try:
        return int(path.read_text(encoding="ascii").strip())
    except (FileNotFoundError, PermissionError, ValueError):
        return None


def write_workflow(path: Path, script_bytes: int) -> None:
    if script_bytes < MINIMUM_SCRIPT_BYTES:
        raise ValueError(f"script must be at least {MINIMUM_SCRIPT_BYTES} bytes")

    # One comment line: '#' + padding + '\n'. Its parsed run-block contents are
    # exactly script_bytes bytes and produce no shellcheck diagnostics.
    body = b"#" + (b"x" * (script_bytes - 2)) + b"\n"
    workflow = b"".join(
        (
            b"on: push\n",
            b"jobs:\n",
            b"  t:\n",
            b"    runs-on: ubuntu-latest\n",
            b"    steps:\n",
            b"      - run: |\n",
            b"          ",
            body,
        )
    )
    _ = path.write_bytes(workflow)


def probe(
    actionlint: str,
    shellcheck: str,
    workflow: Path,
    script_bytes: int,
    timeout: float,
) -> ProbeResult:
    write_workflow(workflow, script_bytes)
    command = [actionlint, "-shellcheck", shellcheck, str(workflow)]

    try:
        process = subprocess.run(
            command,
            capture_output=True,
            check=False,
            timeout=timeout,
        )
    except subprocess.TimeoutExpired:
        return ProbeResult.HUNG

    if process.returncode == 0:
        return ProbeResult.COMPLETED

    stdout = process.stdout.decode(errors="replace").strip()
    stderr = process.stderr.decode(errors="replace").strip()
    details = [
        f"actionlint failed while probing a {script_bytes}-byte run block",
        f"exit status: {process.returncode}",
    ]
    if stdout:
        details.append(f"stdout:\n{stdout}")
    if stderr:
        details.append(f"stderr:\n{stderr}")
    raise ProbeFailure("\n".join(details))


def print_probe(script_bytes: int, result: ProbeResult) -> None:
    label = "ok" if result is ProbeResult.COMPLETED else "HANG"
    print(f"probe {script_bytes:7d} B: {label}", file=sys.stderr)


def main() -> int:
    args = parse_args()
    if args.ceiling <= MINIMUM_SCRIPT_BYTES:
        raise ProbeFailure(
            f"--ceiling must be greater than {MINIMUM_SCRIPT_BYTES} bytes"
        )

    actionlint = resolve_executable(args.actionlint)
    shellcheck = resolve_executable(args.shellcheck)
    capacity = fresh_pipe_capacity()
    soft_limit = linux_pipe_soft_limit()

    print(f"platform:    {platform.system()} {platform.release()} {platform.machine()}")
    print("actionlint:")
    print(
        f"  {version_output(actionlint, '-version').replace(chr(10), chr(10) + '  ')}"
    )
    print("shellcheck:")
    print(
        f"  {version_output(shellcheck, '--version').replace(chr(10), chr(10) + '  ')}"
    )
    if capacity is not None:
        print(f"fresh pipe capacity:        {capacity} bytes")
    else:
        print("fresh pipe capacity:        unavailable on this platform")
    if soft_limit is not None:
        print(f"fs.pipe-user-pages-soft:    {soft_limit} pages")
    print(f"actionlint script prefix:   {len(ACTIONLINT_PREFIX)} bytes")
    print()
    _ = sys.stdout.flush()

    with tempfile.TemporaryDirectory(prefix="actionlint-pipe-repro-") as directory:
        workflow = Path(directory, "workflow.yml")

        low = MINIMUM_SCRIPT_BYTES
        low_result = probe(actionlint, shellcheck, workflow, low, args.timeout)
        print_probe(low, low_result)
        if low_result is ProbeResult.HUNG:
            raise ProbeFailure(
                f"the minimum {low}-byte probe timed out; increase --timeout"
            )

        high = args.ceiling
        high_result = probe(actionlint, shellcheck, workflow, high, args.timeout)
        print_probe(high, high_result)
        if high_result is ProbeResult.COMPLETED:
            message = (
                f"no deadlock at {high} B; this actionlint build is fixed, or "
                f"--ceiling is too low"
            )
            print(message)
            return 0

        while low + 1 < high:
            middle = (low + high) // 2
            result = probe(actionlint, shellcheck, workflow, middle, args.timeout)
            print_probe(middle, result)
            if result is ProbeResult.HUNG:
                high = middle
            else:
                low = middle

    stdin_bytes = high + len(ACTIONLINT_PREFIX)
    print()
    print(f"{low:7d} B  ok")
    print(f"{high:7d} B  HANG")
    print()
    print(f"smallest hanging run block: {high} bytes")
    print(f"with actionlint prefix:      {stdin_bytes} bytes")
    if capacity is not None:
        relation = (
            "exact match"
            if stdin_bytes == capacity
            else f"delta {stdin_bytes - capacity:+d}"
        )
        print(f"fresh pipe capacity:         {capacity} bytes ({relation})")
    return 0


if __name__ == "__main__":
    try:
        raise SystemExit(main())
    except ProbeFailure as error:
        print(f"error: {error}", file=sys.stderr)
        raise SystemExit(2) from error
```

</details>

PR #652, opened 2026-04-20
Fix stale metadata for `actions/create-github-app-token`

### LINKED ISSUE #648: actions/create-github-app-token@v3 inputs changed
Hello @rhysd 

Thanks for all the work you're putting into this very useful project.

We have recently noticed deprecation messages when calling actions/create-github-app-token@v3 as `app-id` should be replaced with `client-id`.

Unfortunately, it seems they introduced this change in v3.1.0: https://github.com/actions/create-github-app-token/blob/v3.1.0/action.yml

But the popular_actions.go file was updated before that (in https://github.com/rhysd/actionlint/commit/875f98c5b6b69304af54871902d690e8c82d8659) and at this point the action was on v3.0.0 which didn't have this change: https://github.com/actions/create-github-app-token/blob/v3.0.0/action.yml

This means the linter fails when calling the action with the non-deprecated setting as `app-id` is wrongly recognized as required and `client-id` is wrongly recognized as invalid.

There has been a number of update to the file since then but I guess the script doesn't attempt to refresh the file as this would be wasteful in most cases.

Would it be possible to regenerate the inputs for this particular action?

Thanks again!

anthonymonori:
`@v3` currently reflects the old `v3.0.0` input shape, which causes false positives when workflows use `client-id` and the current moving `v3` tag.

Example output (before changes in this PR):

```
$ actionlint .github/workflows/cherry-pick.yml
.github/workflows/cherry-pick.yml:91:15: missing input "app-id" which is required by action "actions/create-github-app-token@v3". all required inputs are "app-id", "private-key" [action]
.github/workflows/cherry-pick.yml:93:11: input "client-id" is not defined in action "actions/create-github-app-token@v3". available inputs are "app-id", ..., "private-key", ... [action]
```

[actions/create-github-app-token@v3.1.0 release notes](https://github.com/actions/create-github-app-token/releases/tag/v3.1.0)

This updates the popular actions registry and regenerates `popular_actions.go` so that:
- `actions/create-github-app-token@v3` matches the current upstream `v3` metadata
- `actions/create-github-app-token@v3.0.0` is kept as a separate exact entry for workflows pinned to that version

Fixes #648 

I've checked the CONTRIBUTIONS doc, ran linting, ran `go generate` (there actually was some other unrelated changes for `aws-actions/configure-aws-credentials@v6` that i reverted since it's unrelated to this PR but just an FYI).

Let me know if I missed anything! 🙏 **Or if you would prefer if we don't add the 3.0.0 regression case and just change the 1 line that's matters.**


muzimuzhi:
> Or if you would prefer if we don't add the 3.0.0 regression case and just change the 1 line that's matters.

I think the latter way is adopted in this repo.

BTW, input updates in `popular_actions.go` file can be auto-updated by the [`generate.yaml`](https://github.com/rhysd/actionlint/blob/011a6d15e749bb3f2d771eed9c7aa0e7e3e10ee7/.github/workflows/generate.yaml) workflow, see for example https://github.com/rhysd/actionlint/pull/611.

But recently it failed because new releases of popular actions were detected but they were missing from the `popular_actions.go`:

```
[step] Check new release on GitHub
......
Detected some new releases
2026/04/18 02:04:56 Done detecting new versions in 99 repositories
https://github.com/actions/github-script/tree/v9
https://github.com/actions/upload-pages-artifact/tree/v5
https://github.com/dawidd6/action-download-artifact/tree/v20
https://github.com/softprops/action-gh-release/tree/v3
exit status 2
Error: Process completed with exit code 1.
```
https://github.com/rhysd/actionlint/actions/runs/24594466017/job/71921705206#step:4:207

anthonymonori:
Thanks @muzimuzhi for pointing that out. Was looking around how to address that issue as I ran into it locally, but I didn't see that we have a way to auto-update metadata.

I guess if that workflow would be fixed, we would get a PR that patches this change for v3 anyway? In that case, I'm happy to close this down and go that route. 🙏 

In the meanwhile I also removed the regression case and reduced this PR to the minimal change in [778f197](https://github.com/rhysd/actionlint/pull/652/commits/778f197b563e83efa9723024991672e5d0adf8cb)

smorin:
Looking forward to this one

danedf (approved):

PR #668, opened 2026-05-21
Update `popular_actions.go` for `actions/create-github-app-token@v3`

### LINKED ISSUE #669: False-positive on `actions/create-github-app-token@v3` `client-id` input
## Bug

The bundled v3 schema for [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token)
in `popular_actions.go` is out of date relative to the action's current `action.yml`.

On any workflow that uses the canonical `client-id` input, actionlint emits two
false-positives per `with:` block:

```
missing input "app-id" which is required by action "actions/create-github-app-token@v3"
input "client-id" is not defined in action "actions/create-github-app-token@v3"
```

## Repro

```yaml
# .github/workflows/repro.yml
name: repro
on: push
jobs:
  mint:
    runs-on: ubuntu-latest
    permissions:
      id-token: write
    steps:
      - uses: actions/create-github-app-token@v3
        with:
          client-id: ${{ vars.MY_APP_CLIENT_ID }}
          private-key: ${{ secrets.MY_APP_PRIVATE_KEY }}
```

Run `actionlint .github/workflows/repro.yml` against actionlint v1.7.12 (HEAD as
of this writing). Both diagnostics above fire, but the workflow is valid — it
runs successfully on GitHub Actions, with the action printing a deprecation
warning that recommends `client-id`.

## Root cause

The action's [v3 `action.yml`](https://github.com/actions/create-github-app-token/blob/v3/action.yml)
made `client-id` the canonical input and demoted `app-id` to a deprecated alias:

```yaml
inputs:
  client-id:
    description: 'GitHub App Client ID'
    required: false
  app-id:
    description: "GitHub App ID. Deprecated, use 'client-id' instead."
    required: false
    deprecationMessage: "Use 'client-id' instead."
  private-key:
    description: 'GitHub App private key'
    required: true
  # ... plus new inputs: enterprise, permission-artifact-metadata,
  # permission-attestations, permission-discussions, permission-merge-queues
```

The schema in `popular_actions.go` still marks `app-id` as `required: true` and
doesn't know about `client-id` or the new inputs.

## Why the schema is stale

The weekly [`generate`](.github/workflows/generate.yaml) workflow that refreshes
`popular_actions.go` currently fails on an unrelated upstream 404:

```
request was not successful https://raw.githubusercontent.com/actions-cool/issues-helper/v1/action.yml: 404 Not Found
```

`actions-cool/issues-helper` appears to have moved/deleted its `v1` tag upstream,
so every run of the generator (which iterates the full registry) bails out before
it can re-bake any of the schemas — including this one. So the auto-refresh path
is blocked for everyone in `popular_actions.json` until someone unblocks the
generator.

## Fix

PR #668 narrowly regenerates just this one entry via a single-entry registry,
dodging the broken `actions-cool/issues-helper@v1` fetch. That should make the
false-positives stop landing on consumer repos in the meantime. Separately, the
generator's behaviour on a 404 deserves a fix so a single deleted upstream tag
doesn't gate the whole weekly refresh — but that's a follow-up.

## Workaround for downstream users

Until a release ships with the updated schema, add the following to
`.github/actionlint.yaml`:

```yaml
paths:
  .github/workflows/**/*.{yml,yaml}:
    ignore:
      - 'missing input "app-id" which is required by action "actions/create-github-app-token@'
      - 'input "client-id" is not defined in action "actions/create-github-app-token@'
```

## Version

- actionlint v1.7.12 (HEAD as of 2026-05-21)
- `actions/create-github-app-token@v3` (current upstream HEAD)

esaari:
Closes #669.

## What

Regenerate the v3 entry in `popular_actions.go` for
[`actions/create-github-app-token`](https://github.com/actions/create-github-app-token)
to match its current upstream `action.yml`.

Diff:

- `app-id` — was `required: true`; now `required: false` with `deprecationMessage: "Use 'client-id' instead."`
- `client-id` — added (canonical replacement for `app-id`)
- `enterprise` — added
- `permission-artifact-metadata` — added
- `permission-attestations` — added
- `permission-discussions` — added
- `permission-merge-queues` — added

v1 and v2 schemas were verified against their respective live `action.yml` files
and left untouched — those tags ship `app-id` only and don't need any change.

## Why

The bundled v3 schema still requires `app-id` and doesn't know about `client-id`.
On any workflow using the canonical input, actionlint produces two false-positives
per `with:` block:

```
missing input "app-id" which is required by action "actions/create-github-app-token@v3"
input "client-id" is not defined in action "actions/create-github-app-token@v3"
```

`actions/create-github-app-token@v3` was published 2025-09 with `client-id` as the
canonical input; teams adopting the supported input get this false-positive pair
on every call site. Full repro and root cause in #669.

## Generation method

I tried the full regen first
([`go run ./scripts/generate-popular-actions ./popular_actions.go`](https://github.com/rhysd/actionlint/blob/main/CONTRIBUTING.md#maintain-popular_actionsgo)),
but it currently fails on an unrelated 404:

```
request was not successful https://raw.githubusercontent.com/actions-cool/issues-helper/v1/action.yml: 404 Not Found
exit status 1
```

`actions-cool/issues-helper` appears to have moved/deleted its `v1` tag upstream.
This is also blocking the weekly auto-refresh workflow from updating any entry —
worth a follow-up to make the generator tolerant of upstream 404s, but out of
scope for this PR. I scoped the regen to just the affected action via a
single-entry registry:

```
go run ./scripts/generate-popular-actions \
  -r /tmp/cgat-registry.json -q
```

(where `/tmp/cgat-registry.json` contains only the `actions/create-github-app-token`
slug). The output was then patched into the existing `popular_actions.go` in place.
The resulting diff is byte-for-byte what the full regenerator would produce for
this entry — no manual edits to the generated formatting.

## Verification

- `go build ./...` — pass
- `go test -run TestPopularActionsDataset ./...` — pass
- `go test ./...` — pass


smorin:
Looking forward to this fix

ElDavoo:
Looking backwards, and everywhere around me, for this fix

esaari:
@rhysd Could I please get a review and/or merge on this? Are you planning a release soon?

upalatucci:
cool @esaari thanks for the pr. This is what i was looking for

steveoh:
Anything we can do to help get this merged? Should we be using the fork?

esaari:
@steveoh This has been included in the new active fork: https://github.com/kjanat/actionlint

PR #670, opened 2026-05-25
detect caller/callee permission mismatch in reusable workflow calls

### LINKED ISSUE #552: Actionlint should check permissions passed to reusable workflows (in the same repository)
When calling [reusable workflows](https://docs.github.com/en/actions/how-tos/sharing-automations/reuse-workflows), each reusable workflow requests permissions based on the permissions required for each job. The permissions must be granted by the calling workflow for the workflow to be valid.

If the calling workflow does not grant adequate permissions then GitHub Actions will fail the workflow run with an error like the following.

> The workflow is not valid. .github/workflows/pull_request.tests.yml (Line: 64, Col: 3): Error calling workflow 'ianlewis/repo-template/.github/workflows/workflow_call.zizmor.yml@1ae42c4d4c64e09f3b01302bd64a20af0d82c4bd'. The workflow is requesting 'security-events: write', but is only allowed 'security-events: none'.

Actionlint should check for the correct permissions when the reusable workflow is being called from the same repository via `uses: ./path/to/workflow.yml`. Actionlint already checks for the existence of this file.

SeriousBug:
GitHub validates `permissions:` at workflow load time. When a job in a called reusable workflow declares a scope the calling job does not grant, the run aborts with `startup_failure` before any job can run.

This PR adds a check under the existing `workflow-call` rule that compares each callee job's effective permissions against the caller's effective grant and reports each missing scope.

 When the caller has no `permissions:` block at the workflow level and none on the calling job, the check assumes GitHub's restricted default token (`contents: read` + `packages: read`). This mirrors the repo-level "Workflow permissions" setting (Settings → Actions → General), which actionlint can't read from the workflow file. So unfortunately we need to introduce a config for this, `assume-default-permissions`:

- `restricted` (default) — assume the restricted default
- `permissive` — assume read+write on everything, which effectively skips the check for callers with no permissions block

 Closes #552


aivantg:
Looking forward to this fix! Any info on when it might be merged?

PR #694, opened 2026-06-25
Add support for parallel steps (background, wait, wait-all, cancel, parallel)

### LINKED ISSUE #693: Support for parallel steps
Today GitHub released parallel steps: https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/. This feature introduces 4 new keys:
- `background`
- `wait/wait-all`
- `cancel`
- `parallel`

This is likely to be a quickly adopted feature and would be great if actionlint could be updated to support these new keys and functionality.

devantler:
## Summary

[GitHub released parallel steps on 2026-06-25](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/), adding five new step keys. actionlint currently rejects any workflow that uses them — e.g. `unexpected key "background" for step ...`, and `step must run script with "run" section or run action with "uses" section` for `wait`/`wait-all`/`cancel`/`parallel` steps — which blocks adoption. This teaches the parser about them.

Closes #693

## What the feature looks like

```yaml
steps:
  - name: Start server
    id: server
    run: npm start
    background: true        # run asynchronously, continue immediately
  - wait: server            # wait for one background step (or `wait: [a, b]`)
  - wait-all:               # wait for all preceding background steps
  - cancel: server          # cancel a single background step
  - parallel:               # run a group of steps in parallel
      - run: ./a.sh
      - run: ./b.sh
```

## Changes

Modeled after the spec in [`workflow-syntax.md`](https://github.com/github/docs/blob/main/content/actions/reference/workflows-and-actions/workflow-syntax.md):

- **`background`** (boolean) is parsed as a new optional `*Bool` field on a step (run or action step), mirroring `continue-on-error`.
- **`wait`/`wait-all`**, **`cancel`** and **`parallel`** are modeled as new step execution kinds (`ExecWait`, `ExecCancel`, `ExecParallel`) alongside `ExecRun`/`ExecAction`, so they no longer trip the "step must run …" check.
  - `wait` accepts a single step ID **or a list** (`wait: [a, b]`); `cancel` targets a **single** step ID (a list is reported as a type error, per the spec).
- **`parallel`** parses its grouped steps recursively (reusing `parseSteps`), and the visitor in `pass.go` recurses into them, so **every rule still lints the nested steps**.
- Unknown keys on the new step kinds are reported the same way as for `run`/action steps (e.g. `unexpected key "shell" for step to wait for background steps`).

## Context checking — intentionally deferred (with reason)

actionlint generates its context-availability table (`availability.go`) from GitHub's [`contexts.md`](https://github.com/github/docs/blob/main/content/actions/reference/workflows-and-actions/contexts.md), and that file does **not yet list** `background`/`wait`/`cancel`. So there is no published context-availability to validate `${{ }}` expressions in those fields against, and adding guessed rules now would risk false positives. Expressions are still parsed (so e.g. `background: ${{ ... }}` is accepted), and `name`/`if` on the new step kinds are checked as before. Once these keys are added to `contexts.md`, regenerating `availability.go` plus a small follow-up will cover it.

## Test plan

- New fixtures: `testdata/ok/parallel_steps.yaml` (valid usage of all five keys, incl. an expression-valued `background`) and `testdata/err/parallel_steps.yaml` + `.out` (unexpected keys on wait/cancel/parallel steps; `cancel` with a list; and an invalid step nested in `parallel`, which proves the recursion lints it).
- Updated the existing golden fixtures whose step-key lists now include `background`.
- `go test ./...` is green (1881 tests); `gofmt`/`go vet`/`staticcheck` clean.

---


devantler:
~~It should be good to merge now. Just verified it against the spec one final time :-)~~

Thanks for flagging critical oversights @nikosavola!

parse.go
```diff
@@ -1198,6 +1200,92 @@ func (p *parser) parseStepExecRun(entries []workflowMappingEntry) *ExecRun {
...
+func (p *parser) parseStepExecWait(entries []workflowMappingEntry) *ExecWait {
+	ret := &ExecWait{}
+
+	for _, e := range entries {
+		switch e.id {
+		case "wait":
+			ret.Names = p.parseStringOrStringSequence("wait", e.val, false, false)
+		case "wait-all":
+			ret.All = true
+			ret.AllPos = e.key.Pos
```
nikosavola, line 1220: `case "wait-all"` doesn't record a possible value unlike `wait`. I think we could record and report if a value was given. `wait-all` takes no arguments (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepswait-all) but this kind of user error seems like something actionlint should check for.

nikosavola, line 1220: `wait` and `wait-all` on the same step isn't flagged right now. Both branches can fire, producing Names != nil && All == true. Specifying both is contradictory and a natural lint target.

parse.go
```diff
@@ -1198,6 +1200,92 @@ func (p *parser) parseStepExecRun(entries []workflowMappingEntry) *ExecRun {
...
+// parseStepExecCancel parses a 'cancel' step that cancels running background steps.
+// https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/
+func (p *parser) parseStepExecCancel(entries []workflowMappingEntry) *ExecCancel {
+	ret := &ExecCancel{}
+
+	for _, e := range entries {
+		switch e.id {
+		case "cancel":
+			// The 'cancel' step targets a single background step by its ID (not a list).
+			ret.Name = p.parseString(e.val, false)
```
nikosavola, line 1255: `wait: foo` and `cancel: foo` reference step IDs like `needs:` references job IDs which actionlint validates. Right now an unknown target like `cancel: typo` would pass. Cross-referencing these against the IDs of preceding `background: true` steps would be a high-value check but maybe it could be in another PR.

nikosavola (changes requested):

parse.go
devantler: Good catch. Will fix.

devantler: Good catch. Will fix.

parse.go
devantler: Good catch. Spawning a stacked PR, so it can be reviewed in isolation. It will make the current PR to big to be easily reviewable, so I agree with your suggestion to do it in another PR!

devantler: https://github.com/rhysd/actionlint/pull/695

GodVenn:
@devantler @nikosavola Can this be merged or can I help in any way? We are eager to see this deployed 👼

devantler:
> @devantler @nikosavola Can this be merged or can I help in any way? We are eager to see this deployed 👼

I am waiting for approval. I believe it to be ready.

PR #708, opened 2026-07-25
Accept '+' sign in exponent part of number literals

semx:
The expression lexer accepts a `-` sign in the exponent part of a number literal (`1e-5`) but rejects a `+` sign (`1e+5`), even though both are valid.

The GitHub Actions documentation states that number literals use "any number format supported by JSON", and the lexer's own comment repeats this. The JSON number grammar (RFC 8259 / ECMA-404) allows an optional sign in the exponent:

```
exp = ("e" / "E") ["-" / "+"] 1*DIGIT
```

Go's own `encoding/json` accepts all of these:

```
1e+5     -> 100000
2.99e+2  -> 299
3e+42    -> 3e+42
```

but `actionlint` reports them as syntax errors:

```
got unexpected character '+' while lexing exponent part of float number, expecting '0'..'9'
```

For a linter this is the worst kind of false positive: valid input is flagged as an error.

`lexNum` only checked for `-` after `e`/`E`. This adds `+` alongside it, matching the JSON exponent grammar. `1e5`, `1e-5` and every existing case are unchanged.

New lexer test cases: `1.0e+99`, `1.0E+99`, `3e+42`.

Note: a leading zero in the exponent (`1e05`) is still rejected. That is existing, deliberately-tested behavior (`TestLexExprError` covers `1e01`/`1.0e01`), so this PR leaves it untouched.


semx:
@rhysd, whenever you have a moment — the CI run on this one is still sitting in `action_required`, so there are no checks for you to look at yet. Could you approve the workflow?

For context, the change is two characters in the lexer: `1e+5` and `2.99e+2` are currently rejected while `1e-5` is accepted, because the exponent only allows a `-` sign. Three test cases cover it.

PR #715, opened 2026-08-05
fix: include "type" and "options" in workflow_dispatch input key error

Socialpranker:
_Disclosure: this patch was prepared with AI assistance (Claude Opus 5). I verified the parser behaviour, the test fixtures and the build myself before opening this._

### Problem

`parseWorkflowDispatchEventInput` accepts five keys — `description`, `required`, `default`, `type` (parse.go#L423) and `options` (parse.go#L441) — but its `default` branch reports only three of them:

```go
p.unexpectedKey(e.key, "inputs", []string{"description", "required", "default"})
```

So a user who mistypes a key next to a valid `type:` or `options:` is told those two are not expected, while the parser has just accepted them a few lines above.

The `workflow_call` inputs section already lists all of its keys (`"description", "required", "default", "type"`), so the two sibling sections disagreed on how complete the hint should be.

### Fix

Add the two missing keys to the list. `unexpectedKey` sorts the names via `sortedQuotes`, so the order in the literal does not matter.

Before / after on a workflow that uses `type: choice` + `options:` plus one bad key:

```
- unexpected key "bogus_key" for "inputs" section. expected one of "default", "description", "required"
+ unexpected key "bogus_key" for "inputs" section. expected one of "default", "description", "options", "required", "type"
```

### Testing

- `go build ./...` — clean
- `go test .` — `ok github.com/rhysd/actionlint`
- Two `testdata/err` fixtures encoded the old message and are updated accordingly
- Ran the built linter against a `workflow_dispatch` workflow using `type`/`options` to confirm the new wording

One pre-existing failure, `TestDetectErrorBadRequest` in `scripts/generate-popular-actions`, is unrelated to this change: I reproduced it on a clean checkout without my patch.


Socialpranker:
Friendly ping — this has been open since August 5th, CI green and nothing needed on my end. It's a one-line fix so the error message lists all 5 valid keys instead of 3 — happy to adjust if you'd rather word it differently.

Socialpranker:
Closing this as stale on my side — no reflection on the change itself.

The default branch hasn't seen a commit or a merged PR since April 2026 (the recent pushes are dependabot branches, not `main`). I'd rather not leave a PR sitting in a queue that isn't being processed.

The fix itself remains valid and the branch stays up. If maintenance picks up and this is still useful, feel free to reopen — no need to ask me first.

PR #717, opened 2026-08-05
Fix deadlock when external command stdin exceeds pipe buffer

lawrencecchen:
## Summary

Pass external-command stdin through `cmd.Stdin` instead of synchronously filling an `StdinPipe` before the child process is started.

The current order deadlocks when a workflow `run:` block is larger than the OS pipe buffer: `io.WriteString` blocks waiting for a reader, but the child (for example ShellCheck) is only started later by `cmd.Output()`. Each repeated `actionlint` invocation then remains stuck indefinitely and can consume substantial CPU.

## Fix

- use `strings.NewReader(e.stdin)` as `cmd.Stdin`, allowing `os/exec` to start the child before pumping input
- add a regression test with stdin larger than the pipe buffer

## Verification

- `go test . ./cmd/actionlint`
- the affected 11 KB workflow previously hung indefinitely with four concurrent actionlint processes; the fixed binary completes in 0.18 seconds with ShellCheck enabled
- `actionlint -shellcheck=` also confirmed the hang was isolated to external-command stdin handling

One unrelated network-dependent test in `scripts/generate-popular-actions` returned status 0 instead of its expected synthetic 400 response; core and command tests pass.


lawrencecchen:
Closing as already fixed by #651 / fd33e9f on upstream main. I independently confirmed the same root cause: writing the full payload to StdinPipe before Start deadlocks once the workflow script exceeds the kernel pipe buffer. The upstream fix completes our previously hanging workflow in under 0.2s.

Issue #537, opened 2025-05-09
input "artifact-ids" is not defined in action "actions/download-artifact@v4"


waqaskayani:
The `actions/download-artifact@v4` action has an input field `artifact-ids` as per [documentation](https://github.com/actions/download-artifact?tab=readme-ov-file#inputs), that allows to download artifacts based on their IDs. The linter does not recognizes it and throws this error.
```
input "artifact-ids" is not defined in action "actions/download-artifact@v4". available inputs are "github-token", "merge-multiple", "name", "path", "pattern", "repository", "run-id" [action]
```

This needs to be added to allow:
```yaml
      - uses: actions/download-artifact@v4
        with:
          artifact-ids: ${{ inputs.artifact-ids }}
```

PR #487, opened 2024-12-09
Update `popular_actions.go`

mi-wada:
## What

Follow <https://github.com/rhysd/actionlint/blob/main/CONTRIBUTING.md#maintain-popular_actionsgo>

1. `go run ./scripts/generate-popular-actions -d` and fix `scripts/generate-popular-actions/popular_actions.json`
2. `go generate`

## Why

Looks fails auto generate Actions due to major version release.
https://github.com/rhysd/actionlint/actions/runs/12208933028

Due to not updated `popular_actions.go`, actionlint reports false positive on my repo.
image

PR #718, opened 2026-08-05
Update the popular actions data set

kolyshkin:
## What

Two commits:

1. Remove `actions-cool/issues-helper` from the popular actions data set.
2. Add the new major versions reported by `generate-popular-actions -d`, and regenerate `popular_actions.go`.

Follows [the documented procedure](https://github.com/rhysd/actionlint/blob/main/CONTRIBUTING.md#maintain-popular_actionsgo), same as #487.

## Why

The weekly [`generate`](https://github.com/rhysd/actionlint/actions/workflows/generate.yaml) workflow has failed on **every run since March**, so the data set has not been updated since 2026-03-31. There are two independent causes.

**1. `-d` aborts the job.** Detection works correctly and has been printing the new releases every Saturday, but it exits with status 2 to report them. That fails the step, so `go generate` and the `create-pull-request` steps never run. Adding a new major needs a manual `popular_actions.json` edit, so this is by design — but it also blocks the routine same-major metadata refresh, which is likely why #537, #648, #652, #668 and #669 have all been sitting.

**2. `go generate` fails on its own now.** `actions-cool/issues-helper` was blocked by GitHub for a ToS violation on 2026-05-19. Its `action.yml` 404s at every tag, so generation dies before writing anything:

```
request was not successful https://raw.githubusercontent.com/actions-cool/issues-helper/v1/action.yml: 404 Not Found
```

Hence commit 1 — the generator cannot run at all until that entry is gone.

## User visible effect

Input checking silently does nothing for versions missing from the data set:

```yaml
- uses: actions/setup-go@v5
  with:
    go-versionn: '1.24'   # reported
- uses: actions/setup-go@v7
  with:
    go-versionn: '1.24'   # NOT reported
```

The lookup key is the full `{owner}/{repo}@{ref}` spec, so a stale data set is indistinguishable from an unknown third-party action and the check is skipped. With this PR both are reported. Since `actions/checkout`, `actions/setup-go`, `actions/setup-node`, `actions/setup-python` and `actions/cache` all shipped a new major during the gap, this affects a lot of workflows.

## Details

25 new majors across 23 entries, including `checkout@v7`, `setup-go@v7`, `setup-node@v7`, `setup-python@v7`, `cache@v6` (+ `/save`, `/restore`), `github-script@v9`, `stale@v11`, `codecov-action@v7`. `PopularActions` goes from 218 to 242 entries.

One judgement call worth a look: `actions/add-to-project` is listed with pinned patch versions (`v1.0.1`, `v1.0.2`) because it had no floating `v1` tag. It publishes a floating `v2` now, so I added `v2` / `next: v3` to match the rest of the file rather than continuing with pinned versions. Happy to change it if you'd prefer `v2.0.0`.

## Checks

- `go build ./...`, `go test .`, `gofmt -l`, `staticcheck ./...` all clean.
- `go run ./scripts/generate-popular-actions -d` now reports `No new release was found` (exit 0).
- `actionlint` on its own workflows and on an unrelated repo: no new findings.

`TestDetectErrorBadRequest` in `scripts/generate-popular-actions` fails for me, but it fails on a clean `main` too, so it looks unrelated and I left it alone.

Note this only refills the data set; the workflow will wedge again on the next major release. Happy to follow up separately if you want `-d` moved after `go generate` so routine refreshes still get a PR when a new major is detected.


kolyshkin:
@rhysd PTAL 🙏🏻

kolyshkin:
@rhysd PTAL 🙏🏻

kolyshkin:
OK this is already fixed (in a better way than here) in the  https://github.com/kjanat/actionlint fork:
 - https://github.com/kjanat/actionlint/commit/1c3a18e8
 - https://github.com/kjanat/actionlint/pull/82

PR #724, opened 2026-08-15
Define jobs.<job_id>.result in workflow_call outputs

### LINKED ISSUE #343: `result` property of the `jobs` context throws an error
Greetings!

I've noticed an issue with `actionlint` when it complains `property "result" is not defined in object type {outputs: {}}`

Code:
```yaml
on:
  workflow_call:
    outputs:
      build-result:
        description: Result of the build
        value: ${{ jobs.build.result }}

jobs:
  build:
    runs-on: ubuntu-latest
    timeout-minutes: 1

    steps:
      - name: build
         run: build-command
```

Here the `result` is a property from the `jobs` context: `jobs.<job_id>.result` https://docs.github.com/en/actions/learn-github-actions/contexts#jobs-context

In this case `outputs.build-result` contains the `build` job result which could be `success`, `failure`, `cancelled`, or `skipped`.

Lower in a caller workflow I call another and pass a `boolean` parameter this way and it works
`publish: ${{ needs.build-clients.outputs.build-result == 'success' }}`

Eljees:
Closes #343.

`checkWorkflowCallOutputs` builds the type of the `jobs` context for a reusable workflow with a single property:

    props[n] = NewStrictObjectType(map[string]ExprType{
        "outputs": o,
    })

So a reusable workflow that wants to hand back whether a job succeeded gets:

    .github/workflows/reuse.yml:6:20: property "result" is not defined in object type {outputs: {artifact: string}} [expression]
      |
    6 |         value: ${{ jobs.build.result }}
      |                    ^~~~~~~~~~~~~~~~~

although GitHub Actions does provide `jobs.<job_id>.result` there, and it is one of the more common things a `workflow_call` output carries.

`needs.<job_id>.result` is already typed as a string a hundred and fifty lines up in the same file (`rule_expression.go:865`), so this is the `jobs` context catching up with the `needs` context rather than a new idea.

### Measured

Built from `011a6d1`, three workflows, same binary before and after:

| workflow | before | after |
| --- | --- | --- |
| `${{ jobs.build.result }}` in a `workflow_call` output | error, exit 1 | **silent, exit 0** |
| `${{ jobs.build.reslt }}` — property typo'd | error, exit 1 | error, exit 1 |
| `${{ jobs.build.outputs.nosuch }}` — output does not exist | error, exit 1 | error, exit 1 |

The two controls are the point: the typo still reports, and it now names the type as `{outputs: {artifact: string}; result: string}`, which shows the property really was added rather than the check being loosened. An unknown *output* is still rejected too, so `outputs` stays a strict object.

`needs.<job_id>.result` in an ordinary workflow was accepted before and is still accepted — unchanged.

### Tests

`testdata/ok/workflow_call_job_result.yaml` follows the convention of the 84 fixtures already there. It fails on an unpatched tree, which is what makes it a regression test:

    --- FAIL: TestLinterLintOK/workflow_call_job_result.yaml
        linter_test.go:79: [testdata/ok/workflow_call_job_result.yaml:6:20:
            property "result" is not defined in object type {outputs: {artifact: string}}]

and passes with the change. `go test .` for the main package is green both before and after.

`go test ./...` reports the same 4 ok / 3 FAIL on an untouched checkout as it does with this patch — the failures are the `scripts/generate-*` packages, which fetch from the network. `gofmt -l` and `go vet` are clean.

AI-assisted (LLM used for drafting); the measurements above are mine.


Eljees:
A note on the missing checks here, in case they looked like a problem with the PR: no CI has ever run on it.

GitHub created two check suites for this PR on 15 August; both contain zero runs and both were marked failed exactly 30 days later, on 14 September. That is what an unapproved first-run-approval gate looks like once it lapses: the workflows were never allowed to start, so there is nothing behind those suites to read.

Approving the workflows would produce real results on the current head, and I am happy to rebase onto master first if you prefer a fresh run.

Issue #40, opened 2021-08-24
Local file error when using a remote to get the local file


travis-crowder-kr:
If you checkout a repository to a local path and try to use it, linting will fail with a file not found.

Example:
```
- name: Checkout token action
  uses: actions/checkout@v2
  with: 
    repository: qwerty/create-github-server-token
    path: ./.helpers/create-github-server-token

- name: Generate GitHub Token
  uses: ./.helpers/create-github-server-token
  id: create-token
...
```
```
neither action.yaml nor action.yml is found in directory
```

This error is expected because the file does not exist locally; it gets downloaded in the pipeline.

rhysd:
The use case makes sense. When I implemented the check for local action, I did not assume that the action is checked out from somewhere.

I agree that the check should not raise an error when no action is found in the directory.

Related to #25

rhysd:
Please try `-ignore` option meanwhile.

```
actionlint -ignore 'neither action\.yaml nor action\.yml is found in directory'
```

rhysd:
@travis-crowder-kr BTW, what is the reason not to use the action directly?

```yaml
- name: Generate GitHub Token
  uses: qwerty/create-github-server-token
  id: create-token
```

rhysd:
The fix was released at [v1.6.3](https://github.com/rhysd/actionlint/releases/tag/v1.6.3).

Issue #25, opened 2021-08-06
Support checking out actions


xsc27:
I have a GitHub org with private repos, which includes private actions. While official support for private actions is in GitHub's backlog, I am working around this by using `action/checkout` before `use: ./.github/actions/my-action`. So during linting, I get:

```
test.yaml:13:15: neither action.yaml nor action.yml is found in directory "github/actions/my-action" [action]
   |
13 |       - uses: ./github/actions/my-action
   |               ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

rhysd:
That's not possible because actionlint does not know where the `my-action` directory can be checked out. And modifying user's directory implicitly by linter is not good because private repositories often contain several confidential things. It possibly leads security incidents.

Can you ignore the error when you run without checking out the action?

```
actionlint -ignore 'neither action.yaml nor action.yml is found in directory'
```

This simply ignores the check when it is not available.

xsc27:
I missed there is a way to ignore tests. Thank you.

My original thought was to do something like checking if there is a checkout earlier in the job to that path. But then that would break `Local action inputs validation at with:`.

PR #732, opened 2026-09-01
Support the `$/` self-repository `uses:` syntax

### LINKED ISSUE #711: Support the new `$/` self-repository `uses:` syntax
Hi, and thanks for actionlint — we run it as a required check on our shared workflows repository, and it has caught real mistakes for us.

GitHub announced a new `uses:` syntax today (2026-07-30) that actionlint doesn't recognise yet:

> [Reference same-repository actions with self-repository syntax](https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/)
>
> A `uses:` value that starts with `$/` resolves to your workflow's own repository at the exact commit that is running, with no checkout required.

Per the announcement it works "everywhere the workspace-relative `./` syntax works, including workflow steps, composite action steps, nested composition, and reusable workflow calls". It requires runner `2.336.0` or newer and is currently github.com only.

### Current behaviour

With actionlint 1.7.12:

```yaml
# .github/workflows/example.yml
name: Example
on: workflow_dispatch
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: $/.github/actions/my-action
```

```
example.yml:8:15: specifying action "$/.github/actions/my-action" in invalid format because ref is missing. available formats are "{owner}/{repo}@{ref}" or "{owner}/{repo}/{path}@{ref}" [action]
```

### Expected behaviour

`$/<path>` accepted without a `@ref`, since the absence of a ref is precisely the point — the action resolves to the commit already running.

### Why it matters

The syntax exists to remove a real desynchronisation risk. Referencing an action of the same repository as `owner/repo/path@v1` hardcodes a version inside the workflow, so a caller pinning `workflow.yml@v1.5.0` still loads the `@v1` action, and any push before the tag moves leaves a window where a new workflow runs against an older action.

### Workaround in the meantime

For anyone landing here, `-ignore` scoped to this single message keeps the rest of the linting intact:

```bash
actionlint -ignore 'specifying action "\$/.+" in invalid format because ref is missing'
```

Happy to help test a fix if that's useful. Thanks again for the tool.

ArloL:
Closes #711.

A `uses:` value starting with `$/` resolves against the repository running the workflow at the exact commit running it, so it carries no `@ref` and needs no checkout. GitHub [added the form][changelog] on 2026-07-30, and it is accepted everywhere the workspace-relative `./` form is — workflow steps, composite action steps, nested composition, and reusable workflow calls.

actionlint 1.7.12 reports it as a malformed repository reference:

```
example.yml:8:15: specifying action "$/.github/actions/my-action" in invalid format because ref is missing. available formats are "{owner}/{repo}@{ref}" or "{owner}/{repo}/{path}@{ref}" [action]
```

## Approach

`canonLocalUsesSpec`, in the new `uses.go`, normalises `$/{path}` to `./{path}` and reports whether a spec is local at all. Every site that previously tested `strings.HasPrefix(spec, "./")` now goes through it.

Normalising rather than teaching each call site a second prefix is what the metadata caches need. They are keyed by spec, and `LocalReusableWorkflowCache.WriteWorkflowCallEvent` writes its keys in the `./` form, so a workflow reached as `$/x.yml` would never find metadata written ahead of it.

`isWorkflowCallUsesRepoFormat` needed the prefix as well. It rejects a leading `.` so a local call is never mistaken for a repository one; without the same guard, `$/path/to/x.yml@ref` parsed as owner `$`, repo `path` and was accepted as a valid repository reference rather than reported.

The error message listing accepted reusable-workflow formats now names the third form, and `docs/checks.md` describes it alongside the other action types.

A step written `$/action@v1` is still accepted silently, because no directory named `action@v1` exists and actionlint deliberately does not complain about missing local action directories (#25, #40). That matches the existing behaviour of `./action@v1`, so I left it alone.

## Tests

| File | Covers |
|---|---|
| `uses_test.go` | `canonLocalUsesSpec` against both forms and the near-misses `$action`, `$`, `../action`; plus that both spellings of a path canonicalise identically, which is the property the caches depend on |
| `rule_workflow_call_test.go` | `$/path/to/x.yml` accepted; `$/path/to/x.yml@ref`, `$/` and `$` rejected |
| `testdata/projects/self_repository_uses/` | A project using `$/action` for a local action and `$/workflows/reusable.yaml` for a reusable workflow call |

The project fixture misspells the action's input in a second step, so its expected output asserts a missing-required-input and an undefined-input error. That proves the metadata behind `$/` is resolved rather than silently skipped, which an empty expected output could not distinguish.

`make test` and `make lint` pass — `go vet`, `staticcheck`, `scripts/check-checks`. Two failures on my machine are unrelated and reproduce on a clean checkout of `main`: `TestDetectErrorBadRequest` in `scripts/generate-popular-actions`, and `check-checks` on the pyflakes section when `pyflakes` is absent from `PATH`.

[changelog]: https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/


chuckatkins:
I independently implemented a patch for #711 and found this PR before
submission. The approaches are similar, but my version has a few differences
that may be useful here:

- It keeps the original `$/...` spelling for diagnostics while using the
  normalized `./...` path only for lookup and cache keys.
- It tests actual cache population and cross-syntax lookup in both directions.
- It reuses existing project fixtures and avoids changing unrelated diagnostic
  snapshots.

My implementation is on my fork at commit fd81d12. Would you prefer incorporating those
parts here, or would an alternative PR be useful?

ekbduffy (approved):

esafak:
See if https://github.com/kjanat/actionlint already does it.

katbyte:
Incase anyone else needs it I've created a minimal fork from 1.7.12 at https://github.com/katbyte/actionlint containing just three fixes - #732 ($/ self-repo), #735 (go.yaml.in/yaml/v4 rc.6) and #651 (shellcheck deadlock). It is not maintained, and is just what was required for golangci-lint and zizmor.

For a maintained fork that is moving forward see https://github.com/kjanat/actionlint - unless this repo is revived, I expect that'll be where I switch once it has a bit more history.

PR #735, opened 2026-09-10
Update go.yaml.in/yaml/v4 to v4.0.0-rc.6 to to coexist with golangci

katbyte:
rc.4+ replaced `ParserError`/`UnmarshalError` with `LoadError`/`LoadErrors` carrying a `Mark`, so actionlint no longer compiles in a module that also depends on rc.6 (gosec v2.28.0, and through it golangci-lint v2.13+). Bump to rc.6 and switch the two error paths to `errors.As` on the new types, keeping the message shape.

Two outputs change, golden files updated: syntax error columns are now 1-based like other actionlint positions (`steps: -` reports column 8, not 7), and unknown anchors get a line and column instead of `0:0`.

`go test .`, `go vet`, `gofmt`, `staticcheck` clean. The `scripts/check-checks` and `scripts/generate-popular-actions` tests already fail on main for unrelated reasons.


katbyte:
Incase anyone else needs it I've created a minimal fork from 1.7.12 at https://github.com/katbyte/actionlint containing just three fixes - #732 ($/ self-repo), #735 (go.yaml.in/yaml/v4 rc.6) and #651 (shellcheck deadlock). It is not maintained, and is just what was required for golangci-lint and zizmor.

For a maintained fork that is moving forward see https://github.com/kjanat/actionlint - unless this repo is revived, I expect that'll be where I switch once it has a bit more history.

